PR #80, opened 2025-05-31
fix: remove execinfo for musl systems

Yappaholic:
This patch removes `execinfo.h` header from the source, which is not available on musl libc systems. Removing header does not break the compositor and enables musl users  on different linux distributions (Gentoo, Alpine, Void) to run maomao. Tested on amd64 x86_64 Gentoo Linux with musl libc.


DreamMaoMao:
thanks

PR #88, opened 2025-06-09
feat: support parse keycode in keybind

DreamMaoMao:


DreamMaoMao:
example:

support use code:[num] to repalce keyname or mod key name

```conf
bind=ALT,code:24,killclient,
bind=code:64,code:24,killclient,
bind=code:64+code:133,code:24,killclient,
bind=Alt+code:133,code:24,killclient,


```

PR #103, opened 2025-06-17
refactor: format nix code with alejandra and optimize src path

Yappaholic:
I applied a fix suggested by nix in `nix/default.nix 29:1` file when running `nix develop`:

```bash
You can make Nix evaluate faster and copy fewer files by replacing `./.` with the `self` flake input, or `builtins.path { path = ./.; name = "source"; }`
```

This should improve evaluation speed and remove this warning when fetching and/or using nix flake. `builtins.path` seems available since `nix 2.26`, so this fix won't cause problems for LTS users.

Also formatted `flake.nix` and `nix` directory with [alejandra](https://github.com/kamadorueda/alejandra) because it is commonly used for flake derivations.

`mmsg` has same warning about `builtins.path` in flake. I can send a PR if needed.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,14 +13,13 @@
     };
   };
 
-  outputs =
-    {
-      self,
-      flake-parts,
-      treefmt-nix,
-      ...
-    }@inputs:
-    flake-parts.lib.mkFlake { inherit inputs; } {
+  outputs = {
+    self,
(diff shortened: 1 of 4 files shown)


DreamMaoMao:
thanks,feel free to pr. I'm not familiar with nixos.

PR #105, opened 2025-06-18
feat: update bug_report template and add aarch64 arch to nix flake

Yappaholic:
Maomao compiles for the aarch64 with no issues, and aarch64 is supported in Nix/NixOS.

The result binary: 
```bash
result/bin/maomao: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /nix/store/9vp8z5dhmc194hdf88znglif3wir7i1v-glibc-2.40-66/lib/ld-linux-aarch64.so.1, for GNU/Linux 3.10.0, not stripped
```

Also updated bug report template, since it referenced dwl and dwl patches

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,17 +1,24 @@
 ---
 name: Bug report
-about: Something in dwl isn't working correctly
-title: ''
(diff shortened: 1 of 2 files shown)


DreamMaoMao:
thanks

PR #143, opened 2025-07-01
Fixing a few gramatical errors in the readme

Sqooky:
(I am so sorry for being such a loser but this is the only way I know how to contribute and I really like the project)


DreamMaoMao:
thanks

PR #153, opened 2025-07-06
feat: tablet input support

DreamMaoMao:


DreamMaoMao:
It  won't be merged for the time being because I don't own a tablet device for  test

PR #163, opened 2025-07-07
fix: crash when run maomao in other compositor

DreamMaoMao:


DreamMaoMao:
fix:

https://github.com/DreamMaoMao/maomaowm/issues/154#issuecomment-3046320046

PR #181, opened 2025-07-30
change: change project name to mango

DreamMaoMao:


DreamMaoMao:
As maomao has a lot of ambiguity on the Internet, in order to avoid it being labeled with some strange tags unrelated to technology, I decided to rename it "My Favorite Fruit"

PR #184, opened 2025-08-01
feat: supoort share window in obs

DreamMaoMao:
need wlroots0.20  to support this, not now. the main branch is using wlroots0.19.


DreamMaoMao:
image

image

DreamMaoMao:
if some want try it, here is the mango next version (wlroots-0.20) , without scenefx. 

 git clone -b wl-only https://github.com/DreamMaoMao/mangowc

UnixLudi0:
Am I right in understanding that this will only be included in the main branch when scenefx will use wlroots 0.20?

PR #249, opened 2025-09-20
If focus_cross_monitor is enabled focusdir focuses the monitor if no clients available

levnikmyskin:
This PR enables `focusdir` to focus the monitor in that direction if there are no clients available, and `focus_cross_monitor` is true. I think this makes it more intuitive, requiring only one shortcut (`focusdir`) to focus the other monitor, whether there's a client or not.


DreamMaoMao:
lgtm

PR #251, opened 2025-09-23
Add map input to output to constrain tablet to a device. 

rockerBOO:
Tablets will use all monitors by default, making the usable surface very small. This constrains the tablet to a single monitor, similar to [input `<identifier>` map_to_output `<identifier>`](https://man.archlinux.org/man/sway-input.5#MAPPING_CONFIGURATION) in sway. Might make sense to for the similar behavior if they have multiple tablets on the same machine to map them directly.

- Add map input to output to constrain tablet to a device. 
- Config option tablet_output_name for the monitor to output to 

I can make any changes to variable names or structure. 

Example config: 
```
tablet_output_name=DP-3
```

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -262,6 +262,8 @@ typedef struct {
 
 	char autostart[3][256];
 
+	char *tablet_output_name;
+
 	ConfigTagRule *tag_rules; // 动态数组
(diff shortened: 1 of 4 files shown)


DreamMaoMao:
tablet already support

PR #257, opened 2025-09-25
Comboview: Use first tag_number's layout in combo

argosnothing:
Currently `comboview` is layout-blind. appearing to never change layouts. this corrects that by properly assigning the layout to the first tag in the set, allowing `comboview` to effectively emulate `view` when just Mod+tag_number is pressed, or add on more tags against tag_number1's layout.


DreamMaoMao:
it doesn't need to manually adjust the curtag. It will be uniformly changed in the view and the first tag will be automatically used as the layout tag.

I have corrected it

PR #301, opened 2025-10-18
nix: merge mmsg to mangowc project

### LINKED ISSUE #300: [nix help want]move mmsg into mangowc
To avoid the need to manually fix and submit the hash for mmsg, I have integrated mmsg into mangowc. But I am not familiar with nix configuration. It would be great if someone could help change the nix configuration

ProggerX:
Resolves #300

Basically just removed mmsg from inputs and outputs, because it's one package now.

### DIFF
--- a/nix/nixos-modules.nix
+++ b/nix/nixos-modules.nix
@@ -21,12 +21,7 @@ in {
     environment.systemPackages =
       [
         cfg.package
-      ]
-      ++ (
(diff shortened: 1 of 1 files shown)


ProggerX:
I don't know if it's convenient enough because it's not possible now to override mmsg without overriding all mango in nixos config. But mmsg is part of mangowc now, so I think it's OK.

PR #312, opened 2025-10-19
make hm-module reload on change

JumpIn-Git:
Also someone should move the mmsg nix packaging to this repo.


JumpIn-Git:
moved to #313, sorry

PR #313, opened 2025-10-19
make hm-module reload on change

JumpIn-Git:
someone should test this before its merged @gitOlf

### DIFF
--- a/nix/hm-modules.nix
+++ b/nix/hm-modules.nix
@@ -100,6 +100,12 @@ in {
     xdg.configFile = {
       "mango/config.conf" = lib.mkIf (cfg.settings != "") {
         text = cfg.settings;
+        onChange = ''
+          XDG_CURRENT_DESKTOP=''${XDG_CURRENT_DESKTOP}
+          if [[ $XDG_CURRENT_DESKTOP == "mango" ]]; then
+            mmsg -d reload-config
+          fi
+        '';
       };
       "mango/autostart.sh" = lib.mkIf (cfg.autostart_sh != "") {
         source = autostart_sh;


Yappaholic:
Is `mmsg` referenced in the flake or `default.nix`? I am not sure if the mango module provides mmsg, maybe we can add mmsg to flake inputs

JumpIn-Git:
> Is `mmsg` referenced in the flake or `default.nix`? I am not sure if the mango module provides mmsg, maybe we can add mmsg to flake inputs

the mango package now also exports mmsg

Yappaholic (commented):
Not sure if everyone needs this enabled by default, maybe we could add `reload-on-change` option and then referencing it like `onChange = lib.mkIf (cfg.reload-on-change = true) ...`? Otherwise looks good to me

JumpIn-Git:
> Not sure if everyone needs this enabled by default, maybe we could add `reload-on-change` option and then referencing it like `onChange = lib.mkIf (cfg.reload-on-change = true) ...`? Otherwise looks good to me

@DreamMaoMao

gitOlf:
That is a bad idea. One defines the state with the home manager configuration of the system. You define state! not an operation. The current running system is transfered into the new state after the switch command.
Some configurations do not provide that. One example is the nvidia configuration (you have to reboot to activate the kernel modul) from my point of you that is technical debt.

DreamMaoMao:
> That is a bad idea. One defines the state with the home manager configuration of the system. You define state! not an operation. The current running system is transfered into the new state after the switch command. Some configurations do not provide that. One example is the nvidia configuration (you have to reboot to activate the kernel modul) from my point of you that is technical debt.

Yes, this doesn't seem to be a universal option

JumpIn-Git:
can this be merged?

DreamMaoMao:
> can this be merged?

Since I don't use nix myself, I will wait and see what more nix users think

Yappaholic:
> > can this be merged?
> 
> Since I don't use nix myself, I will wait and see what more nix users think
> 

I'll test it today and see if it works properly, if everything if fine, then it can be safely merged

Yappaholic:
There is an unhadled case where you would run home-manager reconfigure outside of mango, which produces error (because compositor is not running)

argosnothing:
could we simply guard the reload config behind if the compositor is running?

JumpIn-Git:
sure how could i do this

Yappaholic:
> sure how could i do this

You can check IPC socket file under `XDG_RUNTIME_DIR`, or check if `XDG_CURRENT_DESKTOP` is mangowc (if it sets this env var)

JumpIn-Git:
could you make the code snippet for me?

Yappaholic:
> could you make the code snippet for me?

This should work:

```nix
onChange = ''
  if [[ $XDG_CURRENT_DESKTOP == "mango" ]]; then
  ${cfg.package}/bin/mmsg -d reload_config
  fi
'';
```

It is worth noting that the XDG variable can be changed by the user, which will break hotreloading, but in most use cases it will be working as expected

JumpIn-Git:
done

Yappaholic:
> done

Nice, I'll do some minor testing and then we can safely merge it

Yappaholic:
Nope, seems that my idea was wrong:
```bash
/nix/store/1s4al67vsrq2zd2yn3k6442m5inwyfwj-home-manager-generation/activate: line 351: XDG_CURRENT_DESKTOP: unbound variable
```
Maybe the sway home-manager module has a way to reload configuration, should be worth checking

JumpIn-Git:
@DreamMaoMao how should i check if mango is running?

Yappaholic:
So, there are good news and bad news:

Bad: it is possible to get `XDG_CURRENT_DESKTOP` variable with `builtins.getEnv`, but it will make the flake impure, which breaks reproducibility and adds a burden to users(since they would need to pass `--impure` on every rebuild).

Good: instead of checking if the user runs the compositor, we may just ignore error ouput of `mmsg` with:
```nix
onChange = ''
  ${cfg.package}/bin/mmsg -d reload_config 2> /dev/null
'';
```
Hopefully this will work as intended :)

JumpIn-Git:
> So, there are good news and bad news:
> 
> Bad: it is possible to get `XDG_CURRENT_DESKTOP` variable with `builtins.getEnv`, but it will make the flake impure, which breaks reproducibility and adds a burden to users(since they would need to pass `--impure` on every rebuild).
> 
> Good: instead of checking if the user runs the compositor, we may just ignore error ouput of `mmsg` with:
> 
> ```nix
> onChange = ''
>   ${cfg.package}/bin/mmsg -d reload_config 2> /dev/null
> '';
> ```
> 
> Hopefully this will work as intended :)

nope thats not the issue, check if the value even is set https://github.com/nix-community/home-manager/blob/0562fef070a1027325dd4ea10813d64d2c967b39/modules/services/window-managers/hyprland.nix#L411

Yappaholic:
> > So, there are good news and bad news:
> > Bad: it is possible to get `XDG_CURRENT_DESKTOP` variable with `builtins.getEnv`, but it will make the flake impure, which breaks reproducibility and adds a burden to users(since they would need to pass `--impure` on every rebuild).
> > Good: instead of checking if the user runs the compositor, we may just ignore error ouput of `mmsg` with:
> > ```nix
> > onChange = ''
> >   ${cfg.package}/bin/mmsg -d reload_config 2> /dev/null
> > '';
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Hopefully this will work as intended :)
> 
> nope thats not the issue, check if the value even is set https://github.com/nix-community/home-manager/blob/0562fef070a1027325dd4ea10813d64d2c967b39/modules/services/window-managers/hyprland.nix#L411

Then maybe we can do this:
```nix
onChange= ''
  (
  XDG_CURRENT_DESKTOP=''${XDG_CURRENT_DESKTOP}
  if [[ $XDG_CURRENT_DESKTOP == "mango" ]]; then
    ${cfg.package}/bin/mmsg -d reload_config
  fi
 )
'';
```

JumpIn-Git:
sure ill push the change try it, if it doesnt maybe we should try runtime dir

Yappaholic:
> sure ill push the change try it, if it doesnt maybe we should try runtime dir

Both appending to `/dev/null` and hyprland way does not work. We can't try runtime dir, since mango does not create a directory inside `XDG_RUNTIME_DIR`, so there is nothing to check there

JumpIn-Git:
should this be closed? completly forgot about this

Yappaholic:
> should this be closed? completly forgot about this

Probably yes, I still don't know if this can be done purely, and I don't think it's worth getting into again

PR #319, opened 2025-10-20
fix: ensure tag changes happen on correct monitor

WeraPea:
Fixes a bug where an activation request from a client on an inactive monitor would change the tag on the currently focused monitor, instead of the client's monitor. This was caused by the tag being changed before the monitor focus was updated to the client's one.


DreamMaoMao:
should use view_in_mon.

PR #354, opened 2025-10-27
Feature: Implement view_tagmon

reDpz:
## Description
Allows binding a key to view a tag on a specific monitor and also prevents cursor from warping if monitor is already selected.

Example usage:
```
bind=SUPER,1,view_tagmon,1,DP-1
bind=SUPER,2,view_tagmon,2,DP-1
...
bind=SUPER,9,view_tagmon,9,DP-1

bind=ALT,1,view_tagmon,1,HDMI-A-1
bind=ALT,2,view_tagmon,2,HDMI-A-1
...
bind=ALT,9,view_tagmon,9,HDMI-A-1
```

This is just something specific to my workflow that I picked up from using awesomewm and it basically allows me to use multiple monitors without having to manually switch between them before trying to select the tag.


DreamMaoMao:
https://github.com/DreamMaoMao/mangowc/commit/9b5ea5009518cc45cf1e104239321062e9618039

image

 refer to this. Existing functions can be reused without the need to repeatedly implement the same logic

PR #369, opened 2025-10-31
Add smooth focus transition animation for opacity and border

screenfluent:
This adds animated transitions when switching focus between windows. Both window opacity and border color now fade smoothly using cubic bezier easing instead of changing instantly.

The feature uses the existing animation system with two new config options:
- animation_duration_focus (default 400ms)
- animation_curve_focus for easing curve

Changes are backwards compatible. You can disable the animation by setting animation_duration_focus=0 in config file.

Implementation touches 5 files with minimal additions to keep the code clean.


DreamMaoMao:
it has many issues with border color conversion and the existing fade animation. I have fixed it, but more tests are needed to see if there are any potential problems

screenfluent:
Hey sorry for not making it better on start. I was super excited when I found Mango yesterday late evening, but I'm neurodivergent and this hard switch from dimmed windows was holding me back from switching Hyprland -> Mango. It worked for me locally and I thought I will just push it without expecting much. So I really appreciate you picked it up, fixed and already merged. Big kudos to you! 🙏 My T480 works so much better with Mango. Thank you!

PR #437, opened 2025-11-19
feat: add -c arg for config file

VictorTennekes:
As per suggested on discord: https://discord.com/channels/1430889676264177687/1435326422682959923/1435326422682959923

I've added the support for a `-c` flag to inline a config in the shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ cd wlroots
 meson build -Dprefix=/usr
 sudo ninja -C build install
 
-git clone https://github.com/wlrfx/scenefx.git
+git clone -b 0.4.1 https://github.com/wlrfx/scenefx.git
 cd scenefx
 meson build -Dprefix=/usr
 sudo ninja -C build install


argosnothing:
Looks good to me! This will be very useful for instances where we want to use mango as a compositor for things like display managers, where we will likely want a more minimal, but different config from the main mango session one.

DreamMaoMao:
There is a problem with this, that is, when the reload configuration is performed, another configuration is still used

DreamMaoMao:
If you want to specify the configuration for startup, you just need
```
MANGOCONFIG=/usr/folder/path  mango
```
the config path will change to `/usr/folder/path/config.conf`

https://github.com/DreamMaoMao/mangowc/wiki#custom-config-path

VictorTennekes:
Fair point, perhaps like this with a static variable is a better approach?
Maintaining track of the active config file allows for use on `reload_config`.

PR #442, opened 2025-11-20
add portals.conf file

xdavidel:
This fixes portals problems in NixOS. (Mainly screen-casting issues)

### DIFF
--- a/nix/nixos-modules.nix
+++ b/nix/nixos-modules.nix
@@ -26,6 +26,28 @@ in {
     xdg.portal = {
       enable = lib.mkDefault true;
 
+      config = {
+        mango = {
+          default = [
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
image

It is best to add this because wlr portal does not provide this interface, which will cause firefox to wrongly judge and fail to trigger idle suppression

xdavidel:
> image
> It is best to add this because wlr portal does not provide this interface, which will cause firefox to wrongly judge and fail to trigger idle suppression

I'll try to add it and test it for a couple of days before changing the PR.
Changing to draft until than.

xdavidel:
Added.
I also gave it a drove this for the weekend.

PR #448, opened 2025-11-23
feat: support transparent wlr session lock

Rexcrazy804:
adds an option for enabling transparent wlr session lock surface.
See the following video for comparison.

https://github.com/user-attachments/assets/71c885e3-c2db-432e-a582-cc366f3d639b

> There are a couple frames of BLACK screen before the lockscreen's wallpaper kicks in, with this option enabled we get a seamless transition

hyprland implements the same with `session_lock_xray` option

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -339,6 +339,7 @@ typedef struct {
 	int adaptive_sync;
 	int allow_tearing;
 	int allow_shortcuts_inhibit;
+	int transparent_wlr_lock;
 
 	struct xkb_rule_names xkb_rules;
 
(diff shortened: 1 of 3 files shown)


DreamMaoMao:
It seems that lock screen software can be configured to be completely transparent themselves. There is no need for a compositor to implement  this.

Rexcrazy804:
> It seems that lock screen software can be configured to be completely transparent themselves.

Are you sure? From what I am seeing you are creating a locked_bg surface that blocks everything, before the lock screen surface  is drawn. 

I will dig into the lock screen side of things as soon as I can and close the pr appropriately.

DreamMaoMao:
Well, indeed, a colored rectangle was created to block the lock screen

Rexcrazy804:
> It seems that lock screen software can be configured to be completely transparent themselves.

I've tried [quickshell's instructions](https://quickshell.org/docs/v0.2.1/types/Quickshell.Wayland/WlSessionLockSurface/#color) for a transparent lockscreen yielding no better results. If you know of a lock screen software that can be configured to be completely transparent by themselves and works with mangowc without the glaring black frame, it would be of great help.

That said, thank you for your work on mangowc.

DreamMaoMao:
It has no mandatory correlation with wlr. i have rename it to   allow_lock_transparent

PR #453, opened 2025-11-25
feat: add support for variables in config

VictorTennekes:
I've tried to implement a variable option for the configuration file.

This allows for config like so:
```
$terminal = ghostty
bind=SUPER,Return,spawn,$terminal
```

I've been unable to test this locally as I keep running into different issues with packages, would appreciate some help testing this 😄

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('mango', ['c', 'cpp'],
-  version : '0.10.5',
+  version : '0.10.6',
 )
 
 subdir('protocols')


DreamMaoMao:
It should be noted that in the `free_config`, the variable linked list and the memory allocated in the linked list should all released. Currently, there is no such action now.

DreamMaoMao:
another problem, it will make the `$` invalid in `spawn_shell` and `regex match pattern` in the `windowrule` .

VictorTennekes:
Perhaps we can exclude `$` characters that are included in quotes, that would allow for variables in the config while keeping the functionality for the spawn_shell and regex matches in the windowrule.

VictorTennekes:
Having looked a little more at it I guess we can just use env vars for this.

PR #466, opened 2025-11-29
feat: dual-row scroller layout

rasmusq:
I tried to make a new layout that fits better with my 55 inch TV and thought that it could be relevant to implement in the actual build. It functions the same as the scroller layout, but allows for two rows to be an independent scroller layout. 

It supports switching windows between rows and changing the ratio of the two row heights.
The layout name is "dual_scroller". It adds following config options:
```
bind=SUPER+SHIFT,b,togglerow # toggles the window between the two rows
bind=SUPER+SHIFT,j,adjust_dual_scroller_split,0.05 # keybinds to resize the split ratio
bind=SUPER+SHIFT,k,adjust_dual_scroller_split,-0.05 
dual_scroller_default_split_ratio=0.5 # the default ratio at the start of a session
```
It has become my main layout, so it might be relevant for others

I am not very experienced with this kind of development, so suggestions are very welcome.
I also posted a demo on reddit if that is of any use: https://www.reddit.com/r/unixporn/comments/1oyuzgi/mangowc_dual_scroller_and_other_layouts_for/


DreamMaoMao:
it has some problems, such as when switching between this layout and others. When switching back to this layout, the status and position of the Windows may become disordered.

rasmusq:
I'll try to get it more polished when I get some time on my hands. I will keep it up to date with the main branch, as I personally currently use this layout.

PR #476, opened 2025-12-02
No radius/border when single: only apply to tiled windows

### LINKED ISSUE #475: Only apply "No border/gaps when single" to tiled windows
`no_border_when_single` and `no_radius_when_single` should only apply to tiled windows. Currently those options do not apply when a tag has a single tiled window but at least 1 floating one. Also, the radius and border is missing when there is a single floating window on a tag but none tiled.

Samq64:
Resolves #475

This has the side effect of rendering borders on all windows while they are being dragged. 

Because I was already in `check_hit_no_border` I took the opportunity to use early returns although that can be reverted if a single return is preferred or if that doesn't belong with this change.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('mango', ['c', 'cpp'],
-  version : '0.10.10',
+  version : '0.11.0',
 )
 
 subdir('protocols')
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -243,6 +243,8 @@ typedef struct {
 	int32_t idleinhibit_ignore_visible;
 	int32_t sloppyfocus;
 	int32_t warpcursor;
+	int32_t drag_corner;
+	int32_t drag_warp_cursor;
 
 	/* keyboard */
 	int32_t repeat_rate;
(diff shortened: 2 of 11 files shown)


DreamMaoMao:
In fact, he is not a bug. Some people just like this

Samq64:
Do you have some examples? I find this PR's behaviour more consistent especially for those who also use smart gaps.

https://github.com/user-attachments/assets/85729f5c-1ced-4be6-a3dc-fe57d60f2cec


https://github.com/user-attachments/assets/f5947195-67ea-49a2-b277-09a14d2f0fae

DreamMaoMao:
The purpose of the border is to know where the current focus is. 

A single floating window does not need to distinguish focus, so it is also hidden.

More than two Windows should display the border. otherwise, it won't be known where the current focus is.

Samq64:
Thanks for the explanation on why the border when single setting has a different purpose than smart gaps.

A single tiled window behind a single floating one doesn't need a border to distinguish focus either because if the floating one is not focused the one under it almost always is.

DreamMaoMao:
The most recent commit should also include this issue

PR #490, opened 2025-12-07
feat: Move window to center when in scroller layout

Mental-Vortex:
In scroller layout, the current usage of scroller_focus_center forces all windows to center, whereas centerwin_scroller allows for manual centering.

### DIFF
--- a/src/dispatch/bind_define.h
+++ b/src/dispatch/bind_define.h
@@ -713,14 +713,26 @@ int centerwin(const Arg *arg) {
 	Client *c = NULL;
 	c = selmon->sel;
 
-	if (!c || c->isfullscreen)
+	if (!c || c->isfullscreen || c->ismaximizescreen)
 		return 0;
-	if (!c->isfloating)
-		setfloating(c, true);
 
-	c->float_geom = setclient_coordinate_center(c, c->geom, 0, 0);
-	c->iscustomsize = 1;
-	resize(c, c->float_geom, 1);
+	if (c->isfloating) {
+		c->float_geom = setclient_coordinate_center(c, c->geom, 0, 0);
+		c->iscustomsize = 1;
+		resize(c, c->float_geom, 1);
+		return 0;
+	}
+
+	if (!is_scroller_layout(selmon))
+		return 0;
+
+	if (selmon->pertag->ltidxs[selmon->pertag->curtag]->id == SCROLLER) {
+		c->geom.x = selmon->w.x + (selmon->w.width - c->geom.width) / 2;
+	} else {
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
It is necessary to distinguish between horizontal and vertical. Here, you consider all rolling to be horizontal

Mental-Vortex:
> It is necessary to distinguish between horizontal and vertical. Here, you consider all rolling to be horizontal

Thank you, I have already distinguished between the SCROLLER and VERTICAL_SCROLLER cases.
@DreamMaoMao

DreamMaoMao:
You should place the center of the window at the midpoint coordinate. Currently, you have placed the coordinate of the upper left corner at the midpoint coordinate

In addition, arrange should be triggered once to make other Windows correspond to the adjustment position if there are multiple scrolling Windows

Mental-Vortex:
> You should place the center of the window at the midpoint coordinate. Currently, you have placed the coordinate of the upper left corner at the midpoint coordinate
> 
> In addition, arrange should be triggered once to make other Windows correspond to the adjustment position if there are multiple scrolling Windows

Thanks, but I don't understand. Doesn't `c->geom.x = selmon->w.x + (selmon->w.width - c->geom.width) / 2;` calculate the midpoint coordinates in scroller layout? It worked fine in my tests. I did perform an arrange at the end. 
Perhaps I should just make a clean branch commit.

DreamMaoMao:
Oh, yes. The coordinates are already centered. I was mistaken

But the second judgment still needs to proceed
 
 arrange trigger

Mental-Vortex:
> Oh, yes. The coordinates are already centered. I was mistaken
> 
> But the second judgment still needs to proceed
> 
> arrange trigger

Thanks, But I'm still not clear on this. I did perform arrange(selmon, false) at the end.
https://github.com/DreamMaoMao/mangowc/pull/490/files#diff-fc82f9342f9c1caf0fd9ecdba8b04e8b759a9a3363e78efe67241171ebbb3bf9R740

Mental-Vortex:
> You should place the center of the window at the midpoint coordinate. Currently, you have placed the coordinate of the upper left corner at the midpoint coordinate
> 
> In addition, arrange should be triggered once to make other Windows correspond to the adjustment position if there are multiple scrolling Windows

Oh, I think I get it now. You mean we should do it this way?
```c
        if (selmon->visible_scroll_tiling_clients == 1) {
                resize(c, c->geom, 0);
        } else {
                arrange(selmon, false);
        }
```

DreamMaoMao:
There's no problem now. However, it would be better to merge it into centerwin. If it's a scroll tile window, adopt this logic; if it's a floating , keep the original centerwin logic.

PR #501, opened 2025-12-11
feat: restore scroller geom in setfullscreen

Mental-Vortex:
Window geometry not restored after exiting fullscreen in Scroller Layout (e.g., Firefox/mpv), causing the window to move to the screen left.

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -301,7 +301,7 @@ struct Client {
 	/* Must keep these three elements in this order */
 	uint32_t type; /* XDGShell or X11* */
 	struct wlr_box geom, pending, float_geom, animainit_geom,
-		overview_backup_geom, current,
+		overview_backup_geom, current, scroller_geom,
 		drag_begin_geom; /* layout-relative, includes border */
 	Monitor *mon;
 	struct wlr_scene_tree *scene;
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
I don't understand what happened. Could you provide a video

Mental-Vortex:
Okay, this is before,

https://github.com/user-attachments/assets/4fb9b85b-fa32-4d23-aa81-f9ba45964ac4

And this is after.

https://github.com/user-attachments/assets/eb9de21d-a145-4d3b-9dda-08857b60b211

DreamMaoMao:
This behavior should not be carried out here. It should be uniformly judged in the layout code.

Besides, it's hard to determine which situation is reasonable. fullscreen window is also support scroll. Such a simple direct restoration of the recorded position might cause confusion after scrolling through the full-screen window.

Mental-Vortex:
> Besides, it's hard to determine which situation is reasonable. fullscreen window is also support scroll. Such a simple direct restoration of the recorded position might cause confusion after scrolling through the full-screen window.

Thanks, I tried multiple full-screen windows and they worked well. Perhaps we could add an option to enable it？

https://github.com/user-attachments/assets/9ebde360-d27f-4b74-b32e-dca7827d068b

Issue #363, opened 2025-10-31
Windows partially overlap in scroller layout


edpyt:
I am experiencing an issue where windows overlap in the scroller layout:

- zen-browser -> kitty  
   <img width="246" height="261" alt="Image" src="https://github.com/user-attachments/assets/7a831f70-652f-4a2f-a41e-c75d2070a58a" />
- kitty -> kitty  
   <img width="273" height="197" alt="Image" src="https://github.com/user-attachments/assets/e03d094b-4ce2-4066-8569-6594f91c3fc8" />
- Video example:  
   https://github.com/user-attachments/assets/57013bf3-de32-4155-8333-c61cc01319a4

The overlap affects all windows.

In the repo readme, I do not see the same issue in the scroller layout demo video:  
https://github.com/user-attachments/assets/c9bf9415-fad1-4400-bcdc-3ad2d76de85a

My multi-monitor setup is configured as follows:
```conf
monitorrule=HDMI-A-1,0.55,1,scroller,1,1,0,0,1920,1080,60
monitorrule=DP-1,0.55,0,scroller,0,1,1085,0,1920,1080,144
```

And my scroller layout config:
```conf
scroller_structs=20
scroller_default_proportion=0.8
scroller_default_proportion_single=1.0
scroller_prefer_center=1
scroller_proportion_preset=0.5,0.8,1.0
```

Is this expected behavior for scroller layout, or is something misconfigured in my setup?

DreamMaoMao:
Did you install it using the aur package? If so, could you try installing it from the source code

edpyt:
> Did you install it using the aur package? If so, could you try installing it from the source code

I installed it via the AUR package.   
I've also updated to the latest AUR version (mangowc-git r761.3346c8a-1.1), but that didn't resolve the issue.

I will now try installing it from source code and report back.

edpyt:
Building from source fixed the problem. It's all working great now after reboot.

https://github.com/user-attachments/assets/fe2e4935-5c97-4a2d-a199-5f56d86ddbe6

edpyt:
It also fixes https://github.com/DreamMaoMao/mangowc/issues/336 for me

DreamMaoMao:
```bash
meson setup build --buildtype=release
```
Coluld you delete the build directory and then package it using the release method to see if this problem will occur?

I need to rule out whether it's a problem with the build method or the aur packaging tool

edpyt:
[...]

PR #510, opened 2025-12-14
fix: scroller layout window overlap caused by uint

andy-novikov:
When windows in scroller layout need to scroll off-screen to the left, they are positioned with negative x coordinates (e.g., `x=-236`). However, the animation code was using `uint32_t`.

The fix changes x and y position variables from `uint32_t` to `int32_t` in `client_animation_next_tick()`, allowing proper handling of negative off-screen coordinates during animation.

this fixes the issue I've commented: https://github.com/DreamMaoMao/mangowc/issues/363#issuecomment-3649847697


DreamMaoMao:
nice catch

PR #518, opened 2025-12-16
feat: mmsg -W flag to list all windows with details

andy-novikov:
Extends IPC protocol with window_info event and list_windows request to query all windows instead of just the focused one. The new -W flag outputs title, appid, geometry, floating/fullscreen state, and tags for each window.

Usage: mmsg -g -W

Tested it just briefly, for whatever reason my flake doesn't want to pick up these changes and points to `mmsg` from cache.

<details>
<summary>screenshot</summary>
Screenshot 2025-12-16 at 22 05 12
</details>

### DIFF
--- a/mmsg/mmsg.c
+++ b/mmsg/mmsg.c
@@ -42,6 +42,7 @@ static int eflag;
 static int kflag;
 static int bflag;
 static int Aflag;
+static int Wflag;
 
 static uint32_t occ, seltags, total_clients, urg;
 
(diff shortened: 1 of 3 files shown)


DreamMaoMao:
The protocol has a character length limit. If a character is too long, it will be truncated. This protocol is not suitable for outputting a large amount of information

andy-novikov:
Could you please provide more details on protocol limits?
This is sending each window data in a separate message

DreamMaoMao:
This seems a bit limited, and the current output data format is rather chaotic. I prefer to create a separate ipc for these data in the future, just like other synthesizers, and then output a json data table, rather than through wayland's extended protocol

andy-novikov:
that's fine, and I do agree on that. would you be able to outline technical requirements and implementation ideas for that in an issue? I might have a look at that. <3

PR #527, opened 2025-12-20
feat: detect directory in source config

dougEfresh:
TLDR - if  `source`  config  value is a directory, then read the files  and process names that end with `.conf` 

** NOTE ** to reviewer, this is a PoC to invoke discussion . I am  open to change implementation  / design.

## Why? 

*  convenience -  managing configuration becomes basic `cp` and `mv` operations. 
*  ascii file sort -- you can use filenames to sort config loading:  10-bind.conf, 20-monitors.conf. 
*  standard unix/linux pattern. Tools like `sudo` , `pam`, `nginx` use this concept 

## Implementation 

* Add  new function `parse_config_file_or_dir`  which uses  `stat` to detect file or directory , call  `parse_config_file` if file or  `parse_config_dir`  if directory 

## Considerations 

- Use  a new dedicated config field `source-dir`  instead of hijacking `source` config 
-  Add default directory location `$HOME/.config/mango/conf.d`  
-  Or just read by default all `*.conf` files in `$HOME/.config/mango

Themes utilities or shells can simply deposit their own files in this default directory.  I use noctalia and when i enabled /configued mango theme , it wrote into my `$HOME/.config/mango/config.conf`  file . I think it is cleaner to utilize a common "conf" directory. 

Again, open to suggestions

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -1,8 +1,11 @@
 #include <ctype.h>
+#include <dirent.h>
 #include <libgen.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 
 #ifndef SYSCONFDIR
 #define SYSCONFDIR "/etc"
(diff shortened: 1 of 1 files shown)


dougEfresh:
Also, package managers love the "include" directory concept. 
It allows system admins to make changes in `/etc/mango/conf.d`  , while the package owner / distributor controls `/etc/mango/config.conf`  ... i.e. it removes package conflicts with local system changes

PR #531, opened 2025-12-24
added opensuse install instructions

killertofus:


README.md
```diff
@@ -67,9 +75,9 @@ https://github.com/user-attachments/assets/bb83004a-0563-4b48-ad89-6461a9b78b1f
 - libxcb
 
 ## Arch Linux
-The package is in the Arch User Repository and is available for manual download [here](https://aur.archlinux.org/packages/mangowc-git) or through a AUR helper like yay:
+The package is in the Arch User Repository and is availble for manual download [here](https://aur.archlinux.org/packages/mangowc) or through a AUR helper like yay:
```
Yappaholic: This is not related to adding opensuse packaging and `mangowc-git` is still a valid way to install through AUR

README.md
```diff
@@ -95,40 +103,15 @@ emerge --ask --verbose gui-wm/mangowc
 The package is in the third-party Terra repository.
 First, add the [Terra Repository](https://terra.fyralabs.com/).
 
-Then, install the package:
+## opensuse
+The package is in the third-party obs repository.
```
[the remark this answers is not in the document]
killertofus: on opensuse, much like arch/gentoo it has a user repository called OpenBuildService, unlike on arch/gentoo which have builtin user repository support, on opensuse you will have to install opi, with zypper when you opi *foo*  it will search for the package then show then will show you the package in order of native repo -> added repo -> obs 
image

so if you type opi mangowc you can pick the one you want to install

README.md
killertofus: AUR helpers are not supported by Arch Linux. So we should not include a way to install via a [aur helper](https://wiki.archlinux.org/title/AUR_helpers)

Yappaholic: It is still not relevant to opensuse packaging, if you want to "fix" arch installation you need a separate PR
And you also have a typo in word "available"

README.md
```diff
@@ -99,41 +103,16 @@ emerge --ask --verbose gui-wm/mangowc
 The package is in the third-party Terra repository.
 First, add the [Terra Repository](https://terra.fyralabs.com/).
 
-Then, install the package:
-
-```bash
-dnf install mangowc
-```
```
Yappaholic: I haven't noticed, but you broke Fedora Linux installation instructions by accident, which is not good and needs to be fixed

Keep this line as is with `bash` code block for `dnf` and after the `bash` block you can start adding `## OpenSUSE` block

Yappaholic (changes requested):

Yappaholic (changes requested):
At this point, I'd suggest for you to close this PR, remove `opensuse` branch and write doc changes from scratch because it will take too much time for me and you to go over and over again with unnecessary changes to the docs

Also, commits should follow `change-type: change-description` kind of naming, so in your case commit should be named `doc: add opensuse install instrurctions`

PR #544, opened 2025-12-30
Update layout.h

Nytreon:
Using 󰃇 is kinda weird imo.... please use O instead as it makes a lot more sense thanks!


Samq64:
I also changed it as part of #537 but to "OV" instead. I think either one is fine.

PR #556, opened 2026-01-06
layout: add floating layout option

kevichi7:
adds a new layout to the layout registry;
skips arranging when the active layout has no arrange function.


DreamMaoMao:
Image

check tagrule.

kevichi7:
oh thanks

PR #581, opened 2026-01-16
feat(scroller): Add vertical stacking within layout

librepup:
These changes enable the following actions/features:
 - `stack_with_left` = Resize the window left to the currently selected window to 50% of the vertical screen/window size, and move the currently selected window (also resized to half the available size) below the other window in a vertical stack.
 - `unstack` = Move the selected window out of the current stack, integrating it back into the scroller layout at normal size.
 - `revert_size` = Windows remaining in half their size after a stack was dissolved can be resized with this action to take up their original/all available space and re-integrate into the scroller layout.
 
 These actions can be bound to keys in the `config.conf`.
 
 A demonstration of these features can be found in the `demo.mp4` file or on https://codeberg.org/nixpup/MangoWC.

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -4197,8 +4197,8 @@ void pointerfocus(Client *c, struct wlr_surface *surface, double sx, double sy,
 				  uint32_t time) {
 	struct timespec now;
 
-	if (sloppyfocus && c && time && c->scene->node.enabled &&
-		!c->animation.tagining &&
+	if (sloppyfocus && !start_drag_window && c && time &&
+		c->scene->node.enabled && !c->animation.tagining &&
 		(surface != seat->pointer_state.focused_surface) &&
 		!client_is_unmanaged(c) && VISIBLEON(c, c->mon))
 		focusclient(c, 0);


nixpup:
These changes enable the following actions/features:
 - `stack_with_left` = Resize the window left to the currently selected window to 50% of the vertical screen/window size, and move the currently selected window (also resized to half the available size) below the other window in a vertical stack.
 - `unstack` = Move the selected window out of the current stack, integrating it back into the scroller layout at normal size.
 - `revert_size` = Windows remaining in half their size after a stack was dissolved can be resized with this action to take up their original/all available space and re-integrate into the scroller layout.
 
 These actions can be bound to keys in the `config.conf`.
 
 A demonstration of these features can be found in the `demo.mp4` file or on https://codeberg.org/nixpup/MangoWC.

Samq64:
> `revert_size` = Windows remaining in half their size after a stack was dissolved can be resized with this action to take up their original/all available space and re-integrate into the scroller layout.

If you mean the gap within a stack after closing a window then I think it could be done automaticlly.

> `stack_with_left`

Maybe `stack_with_prev` would be a better name assuming it also supports the vertical scrolling layout?

src/mango.c
```diff
@@ -3636,7 +3638,27 @@ void keypressmod(struct wl_listener *listener, void *data) {
...
+	if (c->prev_in_stack || c->next_in_stack) {
+		if (c->prev_in_stack) {
+			c->prev_in_stack->next_in_stack = c->next_in_stack;
+		}
+		if (c->next_in_stack) {
+			c->next_in_stack->prev_in_stack = c->prev_in_stack;
+			if (!c->prev_in_stack) { // c was the head of the stack
+				// The next client becomes the new head, so we need to ensure it's treated as such.
+				// This is implicitly handled by setting its prev_in_stack to NULL.
+			}
```
Samq64: This condition doesn't do anything because `prev_in_stack` is set to `NULL` below it.

DreamMaoMao:
He has many logical problems. I'm not sure if all of them have been fixed
The current configuration is like this


```conf
bind=alt+ctrl,h,scroller_stack,left
bind=alt+ctrl,l,scroller_stack,right
bind=alt+ctrl,u,scroller_unstack,left
bind=alt+ctrl,d,scroller_unstack,right
```

DreamMaoMao:
https://github.com/user-attachments/assets/9a6c9eb1-6911-433d-b95e-62de7e238eac

xtheeq:
Thanks DreamMaoMao for supporting this pr.
Currently `switch_proportion_preset` is working only from the top window from the stack but not from other.

EDIT: fixed in commit: `eb55626`

xtheeq:
> Currently `switch_proportion_preset` is working only from the top window from the stack but not from other.

Same issue for `set_proportion` also, that it only works from the top most window and not other windows in the stack.

EDIT: fixed in commit: `eb55626`

xtheeq:
another bug/issue I have found is that, while moving a window from a stack to another tag, there is no resizing ( or ) adapting the remaining space takes place. 

The video might explain better:

https://github.com/user-attachments/assets/6860d4e9-a089-4b4a-8738-a0538b5030ec


EDIT: fixed in commit `d7a72ce` & `39d2793`

xtheeq:
After the recent commit probably, the window is being resized automatically when scroller_stack left / right is being continuously done. 

Here's what I mean:
https://github.com/user-attachments/assets/d171fac1-b69c-48d0-8f0c-1ffb1f909bc7


EDIT: fixed in commit `59ca7c0` & `df582bd`

DreamMaoMao:
> After the recent commit probably, the window is being resized automatically when scroller_stack left / right is being continuously done
> 
> Here's what I mean:  我的意思是： https://github.com/user-attachments/assets/d171fac1-b69c-48d0-8f0c-1ffb1f909bc7

fixed

xtheeq:
**Suggestion 1:**
Instead of having **two separate binds** for _stack_ and _unstack_, in the discord discussion many preferred for a **single bind**.

To clarify the idea, the image may explain well on the expected behavior. 

Drawing 2026-01-17 18 49 47 excalidraw

Based on current discussion, Option 1 appears to be the preferred approach.

The **two separate binds** and **single bind** can be also made as a config so that user can prefer single or double bind.

There's also a voting made for collecting the preference:
https://discord.com/channels/1430889676264177687/1430894908910075985/1462078547853180929


**Suggestion 2:**
A bind that consume / expel windows from the column it is from.
An representation below:
Drawing 2026-01-17 19 56 50 excalidraw

This is also one of the requested ways to move around windows in scroller window stacking.

Both These suggestions are inspired from the behavior of the wayland compositor `niri` which obviously is know for its scrolling. 

The representation diagram example as an video using niri is below.

https://github.com/user-attachments/assets/436b3267-e473-4c31-a01a-ff30b17097e3

I'm just referring to this because its a wm that does one thing the best and that's widely accepted.
Please don't think I'm making a comparison here and don't take this suggestion negatively.
We love mangowc and the work what you have done. < 3


This is just a suggestion that I'm writing after quite some discussion with the community.

librepup:
Small suggestion/idea to add alongside the vertical stacking: stacker looping.

The concept:
In your `~/.config/mango/config.conf`, there is now a new option called `stacker_loop` that can be set to either `1` (true/enable) or `0` (false/disable). This option, when enabled, allows you to, when invoking either the action to create a stack to the left with no window left of the currently selected one or vice versa for invoking the action to create a stack to the right, to loop around all windows to the other end and create a stack there.

So invoking a stack creation to the left on the selected leftmost window on your desktop, while `stacker_loop=1` is in your config, will move the window all the way to the right of all present windows, and create a new stack with the last window present all the way to the right.

xtheeq:
**Suggestion 1** implementation: https://github.com/nixpup/MangoWC/pull/4
**Suggestion 2** implementation: https://github.com/nixpup/MangoWC/pull/5

If you accept this suggestions, then these prs can be merged.

Demo video with these changes:

https://github.com/user-attachments/assets/c81c3503-a822-4721-836a-907d608eb3a4

DreamMaoMao:
only  scroller_stack now.
```
bind=alt+super+ctrl,Left,scroller_stack,left
bind=alt+super+ctrl,Right,scroller_stack,right
bind=alt+super+ctrl,Up,scroller_stack,up
bind=alt+super+ctrl,Down,scroller_stack,down
```
I won't make it very complicated at the beginning. For subsequent improvements, I can raise a separate issue. I need to confirm whether these requirements are a widely needed requirement

xtheeq:
> only scroller_stack now.
> 
> ```
> bind=alt+super+ctrl,Left,scroller_stack,left
> bind=alt+super+ctrl,Right,scroller_stack,right
> bind=alt+super+ctrl,Up,scroller_stack,up
> bind=alt+super+ctrl,Down,scroller_stack,down
> ```
> 
> I won't make it very complicated at the beginning. For subsequent improvements, I can raise a separate issue. I need to confirm whether these requirements are a widely needed requirement

agree with your decision to not increase the complexity at the beginning.
Here’s a cleaner, more precise version with better flow and clearer terminology, while keeping your intent intact:

---

The purpose of **`scroller_stack up / down`** is unclear.

While testing from commit [94e6827](https://github.com/DreamMaoMao/mangowc/pull/581/commits/94e68278e710eaab82879cc4d5e06044b0f5d5eb), the only observable behavior occurs when a stack already exists.

In that case, only **`scroller_stack down`** produces any visible effect. Regardless of where the bind is triggered (leftmost or rightmost column), invoking it moves the focused window underneath the existing stack. Repeating the bind moves back and forth the bottom window stack and unstack.

`scroller_stack up`, on the other hand, does not appear to have any effect in this scenario.

DreamMaoMao:
> `scroller_stack up`, on the other hand, does not appear to have any effect in this scenario.另一方面， `scroller_stack up` 在这种情况下似乎没有任何效果。

scroller_stack up/down use for vertical scroller layout.

xtheeq:
> > `scroller_stack up`, on the other hand, does not appear to have any effect in this scenario.另一方面， `scroller_stack up` 在这种情况下似乎没有任何效果。
> 
> scroller_stack up/down use for vertical scroller layout.

awesome to have this stacking in vertical scroller too :)
It would be better to change the name to `vertical_scroller_stack` for better indication.

PR #589, opened 2026-01-18
Update default config.conf

TheXnlyGhxst:
Updated the default config to be more organized and have clear explanations for most options.


TheXnlyGhxst:
I did not know before that comments cannot be in-line in the config, so I will update all comments to be above the associated command.

TheXnlyGhxst:
The config is fixed and should be fully functional now. I am using the same file for my current system.

Samq64:
Is preserving the current defaults a goal of this PR? Also, monitor rules were recently changed and the hot corner position is now customizable.

TheXnlyGhxst:
> Is preserving the current defaults a goal of this PR? Also, monitor rules were recently changed and the hot corner position is now customizable.

Not at all, I just happened to put in the PR just before the changes were made, and plan to update it with the new configuration options soon. The purpose of the config is to always be up to date with the most recent mango version and provide a good entry point for new users.

TheXnlyGhxst:
Gonna close this as the config is still in flux with breaking changes happening somewhat frequently. May re-visit when it stabilizes.

Issue #398, opened 2025-11-07
[Feature request] IPC to get cursor position


maskudo:
An IPC to get cursor position would be really useful.
I used it to auto show waybar when cursor reaches screen and hide it otherwise. 
Similarly, it could be useful to implement hot corners for overview mode from any position in the screen other than just the bottom left one currently.

leoVici9:
although i have not tested it , there is this https://github.com/AndreasBackx/waycorner

maskudo:
I couldn't get waycorner to work reliably with waybar. Not sure if that's an issue with waybar or waycorner

DreamMaoMao:
done

PR #592, opened 2026-01-19
feat: ipc to get cursor position

maskudo:
## Description
- added ipc to get cursor position

## Related Issue
https://github.com/DreamMaoMao/mangowc/issues/398

### DIFF
--- a/mmsg/mmsg.c
+++ b/mmsg/mmsg.c
@@ -42,6 +42,7 @@ static int32_t eflag;
 static int32_t kflag;
 static int32_t bflag;
 static int32_t Aflag;
+static int32_t pflag;
 
 static uint32_t occ, seltags, total_clients, urg;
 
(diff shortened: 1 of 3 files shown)


DreamMaoMao:
This does not meet your requirements because it cannot monitor mouse events and update them in real time. If an ipc event is triggered every time the mouse moves, the load would be too heavy

On the client side, there are ways to detect whether the mouse has entered a certain area, such as quickshell and wayedge. These implementations from the client's perspective have the lowest cost

To sum up, this function should be detected by the client itself. Forcing support on the compositor is too heavy.

maskudo:
It fits my needs of getting cursor position on demand. I just poll the cursor position with a shell script as watching it for every change would be too expensive an operation.

I don't know about its general applicability for all users though.

DreamMaoMao:
be replaced by

https://github.com/mangowm/mango/pull/1078

Issue #593, opened 2026-01-19
Possibility to configure a bind with SUPER key only


rottakore:
I'm trying to do:
bind=NONE,Super_L,spawn,(command)
bind=SUPER,,spawn,(command)
bind=,Super_L,spawn,(command)

And none of the above, or similar, works as intended, which is running a command by only pressing the SUPER key.

vaskark:
I know you probably want bind**r** but this is exactly what I'm trying to do for the Super key.

Baba-Is-Win:
@rottakore Super,Super_L,spawn,[command] or None,Super_L,spawn,[command]   should work init?

edit: oh no it doesnt, interesting

k0Yq:
> I know you probably want bind**r** but this is exactly what I'm trying to do for the Super key.

Yeah, bindr is better because it only registers the command when you release the key.

Baba-Is-Win:
yeah this would be very convenient

DreamMaoMao:
```
bind=none,Super_R,spawn,foot
```

done.

MichaelThomas0721:
On version 0.12.2, the super key cannot be used alone without breaking keybinds that use the super key.
Examples:
`bind=none,SUPER_R,spawn_shell,<command>` runs whenever SUPER is pressed, this means super keybinds using the super key will trigger this one when run.
`bindr=SUPER,SUPER_R,spawn_shell,<command>` runs on the super release, this also means super keybinds using the super key will trigger this one when run.
`bind=SUPER,SUPER_R,spawn_shell,<command>` doesn't do anything
`bindr=none,SUPER_R,spawn_shell,<command>` doesn't do anything

Are there any plans to be able to use the super key by itself without triggering other keybinds that use the super key?

Baba-Is-Win:
<https://github.com/mangowm/mango/issues/689#issuecomment-3993272932>

PR #596, opened 2026-01-21
Allow mouse side buttons to be bound without modifier aswell as allow SUPER to be bound only on release bind

DizzyJune:
as title says, allows mouse side buttons to be bound without modifier. only the two common ones that most mice have, no support for the 12 buttons ones and such, also allows SUPER to be bound via release binds (bindr). must be bound as following: `bind = SUPER,Super_L,DISPATCHER,OPTIONS` Super_R should work as well. 

this is a small commit directly targeted towards #593, as well as just adds the side buttons as something ive found helpful and have been consistently patching in myself.


DreamMaoMao:
I think this contains two completely unrelated requests. I disabled the mod button in the code for other reasons and it has nothing to do with the binding of the button.

PR #622, opened 2026-02-01
Validate config in home-manager

Tomate0613:
Will make the derivation fail if the config is incorrect

### DIFF
--- a/nix/hm-modules.nix
+++ b/nix/hm-modules.nix
@@ -17,12 +17,7 @@ self: {
     ${cfg.settings}
     EOF
 
-    output=$(${cfg.package}/bin/mango -c "$out" -p 2>&1 || true)
-
-    if echo "$output" | grep -Fq '[ERROR]:'; then
-      echo "$output"
-      exit 1
-    fi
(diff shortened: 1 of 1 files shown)


nix/hm-modules.nix
```diff
@@ -12,6 +12,18 @@ self: {
...
   '';
+  validatedConfig = pkgs.runCommand "mango-config.conf" { } ''
+    cat > "$out" <<'EOF'
+    ${cfg.settings}
+    EOF
+
+    output=$(${cfg.package}/bin/mango -c "$out" -p 2>&1 || true)
+
+    if echo "$output" | grep -Fq '[ERROR]:'; then
+      echo "$output"
```
iynaix:
```suggestion
    ${cfg.package}/bin/mango -c "$out" -p
```

iynaix: -p now has a non-zero exit code on validation errors after #639

Tomate0613:
I think I might have rebased incorrectly

Tomate0613:
Going to close this pr and open a new one rather than messing with the git history

Tomate0613:
#655

PR #635, opened 2026-02-05
feat(Dwindle): Hyprland-like dwindle tiling layout.

Yujonpradhananga:
This PR implements the Dwindle layout, a popular binary tree-based tiling algorithm similar to the one found in Hyprland.

Implemented [dwindle()]: Added the core logic which builds a binary tree from the client list.

Added the `DW` (Dwindle) layout to the registry in 

How it works
- The layout recursively splits the container of the last active window.
- Dynamic Splitting: Checks the aspect ratio of the node to determine whether to split horizontally or vertically (e.g., if `width > height`, it splits side-by-side).
-Gap Support: Fully supports existing gap settings (`gappih`, `gappiv`, center gaps) and smart gaps.

- Verified by switching to the `DW` layout.
- Confirmed that new windows correctly split the available space of the previous window.

### DIFF
--- a/src/layout/horizontal.h
+++ b/src/layout/horizontal.h
@@ -930,3 +930,148 @@ void tgmix(Monitor *m) {
 		return;
 	}
 }
+
+// Dwindle layout
+// Each new window splits the space in half, alternating between
+// horizontal and vertical splits based on aspect ratio
+typedef struct DwindleNode {
+	struct wlr_box box;
+	Client *client;				 // NULL if internal node
+	struct DwindleNode *children[2]; // NULL for leaf nodes
+	bool split_horizontal;		 // true = left/right, false = top/bottom
+} DwindleNode;
+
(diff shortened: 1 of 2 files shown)


DreamMaoMao:
This layout existed before, but I removed it. I think compared with the tile layout, it doesn't offer any additional practicality.

Yujonpradhananga:
Thanks for the feedback!  The main reason I looked to bring this layout was due to the familarity with it from hyprland and in wider monitors it saves me from mannually adjusting the sizes.
Maybe implementing windows spawning based on the mouse cursor much like hyprland and following the binary splitting behavior might be better suited here

DreamMaoMao:
The cutting layout based on the current focus window is more meaningful, but it comes at a high cost to implement

Yujonpradhananga:
all good. I realized the deck layout is named as K, is there a reason to naming it K and not D?

PR #642, opened 2026-02-08
feat: add a source-optional keyword

iynaix:
Adds a `source-optional` keyword for adding a conf file that won't error if it doesn't exist.

The motivation is for dynamic theme / color changing systems like matugen that can generate an optional separate color .conf.

I can't find where the docs are located; or I would added this to the docs as well.


DreamMaoMao:
image

The names here are inconsistent.

iynaix:
Whoops, missed that one. Fixed!

PR #645, opened 2026-02-08
support isfakefullscreen as windowrule property

jampe:
this pr adds isfakefullscreen as windowrule property. for example:

```
windowrule=appid:code,isfakefullscreen:1
```


src/mango.c
```diff
@@ -1484,6 +1483,10 @@ void applyrules(Client *c) {
 
 	setfullscreen(c, fullscreen_state_backup);
 
+	if (c->isfakefullscreen) {
```
jampe: @DreamMaoMao  I hope this location is fine as I dont really understand the surrounding code that well...

Issue #649, opened 2026-02-09
One shot / startup window rules


iynaix:
Goal: On startup, I want to open certain programs on certain tags.

Currently, as far as I can tell, there isn't really a way to achieve this. (Hopefully I'm missing something obvious.)

The closest I can get is to create window rules for applications to open on certain tags. However, these rules are permanent and apply to all future instances of the application as well, e.g. all future browser windows are now also stuck on tag 1. This can be mitigated somewhat by opening the application with a custom window app-id or title, but not all applications support this, e.g. chrome does not set an app-id for incognito windows.

Prior Art:

Hyprland does this by allowing rules to be specified inline in exec-once, e.g.

```
exec-once = [workspace 4 silent] kitty
```

Niri does this with an `at-startup` rule, that only matches during the first 60 seconds after starting niri, e.g.

```kdl
window-rule {
    match app-id="^kitty$"  at-startup=true
    open-on-workspace "4"
}
```

BSPWM does this with one-shot rules that only apply to the next window matching the pattern, then the rule is deleted, e.g.

```sh
bspc rule -a Kitty -o desktop='^4'
```

I think hyprland's approach is the most elegant, since the 60s that niri uses for at-startup is arbitrary, but either approach is fine really.

DreamMaoMao:
It is difficult to associate the executed program with the rules. The simplest way is to set a one-time rule, allowing only one rule to match once.

jabuxas:
im using a custom version of mango while this is not merged, the feature is just too useful

DreamMaoMao:
https://github.com/mangowm/mango/commit/c9e69f9aa1dc3e242042036fb0739ae7b9140486

refer to this.

PR #654, opened 2026-02-11
feat: atstartup rule that only applies within 60s of startup

iynaix:
Add atstartup rule that only applies within the first 60s of mango starting, the motivation is for opening applications on certain tags on boot, outlined in #649.

I chose to do the startup rule since it seems to be the easiest one to implement; I don't see a simple way of implementing the generic one-shot rule.

Ideally, I would update the docs too, but I can't find a public repo for those. :sweat_smile:

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -78,6 +78,7 @@ typedef struct {
 	int32_t ignore_maximize;
 	int32_t ignore_minimize;
 	int32_t isnosizehint;
+	int32_t atstartup;
 	int32_t indleinhibit_when_focus;
 	char *monitor;
 	int32_t offsetx;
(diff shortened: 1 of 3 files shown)


ignition-ctrl:
Forgive if I'm misunderstanding, but if you set atstartup as default `-1`,  
in `applyRules()`, your check `if (r->atstartup && get_now_in_ms() - startup_time > 60 * 1000)` always resolves to true after 60secs, and all rules get skipped. Since `-1` is truthy and `-1 && 1 = true`. If I'm not missing something obvious.

iynaix:
Yes, that was a stupid mistake, I forgot it isn't a boolean.

DreamMaoMao:
https://github.com/mangowm/mango/commit/c9e69f9aa1dc3e242042036fb0739ae7b9140486

refer this

PR #655, opened 2026-02-11
validate config in home-manager

Tomate0613:
Will make the derivation fail if the config is incorrect


Tomate0613:
This is no longer required, since #667 got merged

PR #664, opened 2026-02-15
fix: Warning: Require Meson version >= 0.56.0

squassina:
fix for the message: 

`Warnning: mangowc-git/protocols/meson.build:3: WARNING: Project does not target a minimum version but uses feature deprecated since '0.56.0': dependency.get_pkgconfig_variable. use dependency.get_variable(pkgconfig : ...) instead`

This is minor but was bothering me.


DreamMaoMao:
This doesn't seem to solve the problem

PR #685, opened 2026-02-19
Add nix option to enable/disable adding to session manager

jtliang24:
I added an option to disable adding mangowc to the displaymanager.sessionPackages.

I use UWSM and the extra non-UWSM session clutters the sessions menu, so having the option to disable it would be useful.

### DIFF
--- a/nix/nixos-modules.nix
+++ b/nix/nixos-modules.nix
@@ -9,6 +9,10 @@ in {
   options = {
     programs.mango = {
       enable = lib.mkEnableOption "mango, a wayland compositor based on dwl";
+      addLoginEntry = lib.mkEnableOption {
+        default = true;
+        description = "Whether to add a login entry to the display manager for mango";
+      };
       package = lib.mkOption {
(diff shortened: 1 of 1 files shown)


nix/nixos-modules.nix
```diff
@@ -9,6 +9,10 @@ in {
   options = {
     programs.mango = {
       enable = lib.mkEnableOption "mango, a wayland compositor based on dwl";
+      sessionPackages = lib.mkEnableOption {
```
Yappaholic: `sessionPackages` is ambiguous, since user would expect a list of packages instead of a boolean, something like `addLoginEntry` would be better

jtliang24: I have changed it to addLoginEntry

DreamMaoMao:
I'm planning to roll back this submission. It seems that he has disrupted many people's systems. Its default parameter doesn't seem to work. He asks everyone to display adding this field to enable it.

PR #693, opened 2026-02-20
feat: allow-none-key-name

### LINKED ISSUE #689: Allow "none" as key name.
It would be extremely useful to be able to use "none" as the key name, making stuff like this possible:
```
bindr=SUPER,none,toggleoverview
```
Which would mean something like "If the modifier key is released without having triggered any other keybind, then..."

Yujonpradhananga:
## Summary:
This PR implements true modifier-only keybindings (e.g., bindr=SUPER,none,toggleovervie).
It allows an action to fire when a modifier key is released, but only if no other key was pressed during the hold.
Introduces mod_key_used and prev_mods to track state. If you press SUPER+R to reload, the solo SUPER release won't accidentally trigger a "none" binding.
The config parser now recognizes the "none" key. It also includes safety logic that automatically forces release mode (bindr) for these bindings and issues a warning if the user accidentally uses the standard

Closes #689

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -20,7 +20,7 @@
 
 enum { NUM_TYPE_MINUS, NUM_TYPE_PLUS, NUM_TYPE_DEFAULT };
 
-enum { KEY_TYPE_CODE, KEY_TYPE_SYM };
+enum { KEY_TYPE_CODE, KEY_TYPE_SYM, KEY_TYPE_NONE };
 
 typedef struct {
 	uint32_t keycode1;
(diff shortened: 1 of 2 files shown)


demperor-music:
The `none` keybinds seem to trigger if trying to activate a shortcut involving an additional modifier.
Example: Super,none,toggleoverview activates when pressing SUPER+SHIFT if you start the chord by pressing SUPER rather than SHIFT

Yujonpradhananga:
Oh my bad, I fixed the bug in f0fe874, now it should work as intended @demperor-music

mariokhz:
thank you!!!

mariokhz:
if you are using bind=SUPER,none, doing SUPER+CLICK to move a window will activate the bind once SUPER is released

Yujonpradhananga:
Oh I didnt think about the mouse click behaviour at all as I have never used it , but I've added the fix at 6236adb 
I've squashed the commits too for better readability
could you try it out @mariokhz @demperor-music

Yujonpradhananga:
@DreamMaoMao  Can this be merged  Maomao?

demperor-music:
The fix works just fine on my machine

demperor-music:
The currently documented way of having modifier-only keybinds (e.g. bindr=SUPER,SUPER_L) doesn't solve the same issue that this pull request would solve, as activating any other shortcut triggers that bindr. @DreamMaoMao Can this be merged?

PR #694, opened 2026-02-20
docs: fix guix installation instructions

Mujk:
- Rename GuixSD to Guix System (the distro was renamed in 2019)

- Add branch "main" to channel instructions. Without it, Guix tries to find a "master" branch

- Fix declarative installation (not needed if mangowc is installed via "guix install") 
  - change cons to cons*, so it's a list not a pair
  - add missing "-git" to mangowc


Yappaholic:
Thank you for the fix, I wrote these docs mostly from memory :D
LGTM

PR #696, opened 2026-02-22
made the arch installation better

killertofus:


Yappaholic:
This makes arch docs less intuitive because of removed AUR helper example
Also removing `yay` command from the list of personal DreamMaoMao dependencies for his/her dotfiles makes it worse since other distributions have their own package names

ghost:
AUR helpers are not supported by Arch Linux. so we should not include a way to install via a [aur helper](https://wiki.archlinux.org/title/AUR_helpers)

you Shouldn't use a aur helper as a crutch, meaning you shouldn't use a aur helper for everything, you should instead pacman -S the packages in the native repos, then makepkg the ones in the aur, or use a aur helper

Samq64:
Then shouldn't there at least be instructions on how to build both `mangowc` and `scenefx0.4` with makepkg in the readme?

Also I agree that not mentioning which distro the dot file dependencies are for is confusing especially with some of them being fairly niche.

ghost:
added

Yappaholic (changes requested):
Breaking markdown is not good because headers need to have significance in the docs, especially using first header

Also, **SQUASH YOUR COMMITS**, the docs PR needs to be a single commit and it's not that hard or demanding to soft reset your commits and then force-push to the repo

PR #698, opened 2026-02-23
added opensuse install instructions

killertofus:
for those who are on opensuse


Yappaholic:
PR needs to be a single commit without the `merge` commit, also commit naming does not follow `<commit-type>: <commit description>`, you can check previous repo commits related to packaging and see how it is done

Yappaholic (changes requested):
You are trying to push commit that changes code outside of README, not good, you should learn more about git workflow on GitHub before attempting to send PR, because this PR could bork the whole repo

PR #720, opened 2026-03-01
Revert "Add nix option to enable/disable adding to session manager"

DreamMaoMao:
Reverts mangowm/mango#685


jtliang24:
I proposed #723 which correctly sets the default now.

PR #725, opened 2026-03-02
Tablet Support With No Bugs I am Aware of

Sqooky:
@WeraPea provided a patch fixing the only bug for my fork with tablet support, and with that bug down, after an hour of testing I haven't found any bugs. This should be suitable to be merged upstream with no issues.

Please give all credit to @WeraPea, I owe them greatly


Yappaholic:
@Sqooky, there is an error in your PR, since you probably didn't start with a clean pull from upstream and applied patch manually, hence why your diff is 6K+ lines long.

You should revert any changes, update your fork by pulling from upstream without merging, apply patch manually and then submit a PR

PR #728, opened 2026-03-02
♻️ refactor: split switch_layout into next_layout and previous_layout…

EslamMohamed365:
… functions

- Add next_layout() to cycle forward through layouts
- Add previous_layout() to cycle backward through layouts
- Keep switch_layout() as backward-compatible alias to next_layout()
- Update config.conf with example keybinds for both new functions


DreamMaoMao:
A parameter should be added to this schedule instead of splitting it into two schedules.

PR #734, opened 2026-03-05
opt: support restore size per when master change

DreamMaoMao:


DreamMaoMao:
It might be able to solve this problem. It hasn't been tested yet. I'm not sure if any potential problems have been introduced

PR #738, opened 2026-03-06
mmsg-all-windows

torusJKL:
This PR adds the possibility to retrieve the geometry of all visible windows on all connected monitors.

Example output:

```
mmsg -a
eDP-1 1,26 1055x1053 "emacs" "mango.c - GNU Emacs at Lati5300U"
eDP-1 1057,26 862x1053 "foot" "foot"
HDMI-A-1 1921,26 1407x1573 "kitty" "s cat README.md"
HDMI-A-1 3329,26 1150x786 "foot" "foot"
HDMI-A-1 3329,813 1150x786 "emacs" "*scratch* - GNU Emacs at Lati5300U"
```

This allows a  screen grabbing tool like grim with slurp to present the user with areas of the individual windows so that he can hover over any window and take a screenshot of it without drawing a region manually.

### DIFF
--- a/mmsg/mmsg.c
+++ b/mmsg/mmsg.c
@@ -42,6 +42,7 @@ static int32_t eflag;
 static int32_t kflag;
 static int32_t bflag;
 static int32_t Aflag;
+static int32_t aflag;
 
 static uint32_t occ, seltags, total_clients, urg;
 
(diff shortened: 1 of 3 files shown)


DreamMaoMao:
wayland's extended protocol has a string length limit. The correct approach is to completely abandon dwl ipc and use an independent ipc socket for communication

In addition, the next version of wlroots supports direct window capture by the client and does not require such an operation

torusJKL:
Thanks for the comment.
I'm closing the PR.

Let's see what the new wlroots will bring to the table.

PR #739, opened 2026-03-06
Make number of tags configurable (tag_count)

gplusplus314:
This PR introduces the `tag_count` configuration variable. When the variable is missing from the config, it defaults to 9, which is the previously hardcoded value. This is implemented such that Mango must be restarted for this configuration variable to take effect. It was done this way because the size and scope of the PR would increase dramatically if it was supported by a hot configuration reload.

It works out of the box with Dank Material Shell - the IPC still works and DMS even shows and switches between the correct number of tags.


hboetes:
I've set `tag_count = 10` in my config, but I get the works space 2 and 4 at the same time when I go to workspace 10 with `bind=SUPER,0,view,10,0` Super+0

image

hboetes:
I'm so impatient.  😅

```diff
Binary files 1/.git/index and 2/.git/index differ
diff -u -ruw 1/src/config/parse_config.h 2/src/config/parse_config.h
--- 1/src/config/parse_config.h	2026-03-06 17:38:04.940543509 +0100
+++ 2/src/config/parse_config.h	2026-03-06 17:38:17.577672294 +0100
@@ -1578,6 +1578,7 @@
 		config->default_nmaster = atoi(value);
 	} else if (strcmp(key, "tag_count") == 0) {
 		config->tag_count = CLAMP_INT(atoi(value), 1, 32);
+		tag_count = config->tag_count;
 	} else if (strcmp(key, "center_master_overspread") == 0) {
 		config->center_master_overspread = atoi(value);
 	} else if (strcmp(key, "center_when_single_stack") == 0) {
Only in 1/src/config: parse_config.h.orig
Only in 1/src/dispatch: bind_define.h.orig
diff -u -ruw 1/src/ext-protocol/ext-workspace.h 2/src/ext-protocol/ext-workspace.h
--- 1/src/ext-protocol/ext-workspace.h	2026-03-06 17:38:04.941927965 +0100
+++ 2/src/ext-protocol/ext-workspace.h	2026-03-06 17:38:17.580583016 +0100
@@ -106,7 +106,7 @@
 }
 
 static void add_workspace_by_tag(int32_t tag, Monitor *m) {
-	char name_buf[4];
+	char name_buf[12];
 	const char *name = get_name_from_tag(tag, name_buf, sizeof(name_buf));
 
 	struct workspace *workspace = ecalloc(1, sizeof(*workspace));

```

gplusplus314:
Hey @hboetes , did you restart the compositor after setting the `tag_count` config var? Also, are you sure you didn't already have another bind in your config for `super,0` to enter overview mode?  I'm not able to reproduce your issue. That said, your patch did lead to some changes; thanks for that. See my latest commit.

This has been stable for me, so I'm going to mark this PR as ready for review.

hboetes:
Thanks for incorporating the fix in the first hunk of my patch! One small note on `name_buf`: the `CLAMP_INT` in the `snprintf` call doesn't actually silence the `-Wformat-truncation` warning, because the compiler sees the theoretical range of a `uint32_t` (up to 10 digits), not the runtime-clamped value. The buffer just needs to be 12 bytes (`10 digits + NUL + 1 spare`) to make both the compiler and the logic correct. The `CLAMP_INT` can then be dropped as it's redundant, given the earlier `tag > tag_count` guard.

gplusplus314:
I didn't get that warning from either GCC or Clang, but I was generally unhappy with the `get_name_from_tag` function, so I took a pass at disambiguating the word `tag` (clarified whether it's a number in `[1, tag_count]`) and hardened the `get_name_from_tag` (now renamed to `get_name_from_tag_number` for clarity) function. The new signature should serve the needs of named workspaces in the future, at least while workspaces are still mapped to tags.

gplusplus314:
I took another pass at making the primitives for representing a tag more consistent to avoid future UB.

I also began simplifying the `CLAMP` macros. @DreamMaoMao - let me know what you think of these and I can either do the full refactor in this PR or do a separate one.

hboetes:
At the first restart I changed from 10 to 5, correct reply.

```
724:00:02:55.664 [INFO] [../src/config/parse_config.h:1605] tag_count change requires restart (current: 10, requested: 5)
```

And at this restart I changed the number of tags to -1, and nothing spectacular happened. Also good.

```
743:00:04:05.811 [INFO] [../src/config/parse_config.h:1605] tag_count change requires restart (current: 10, requested: 1)
```

azrtareef:
PLEASE ADD THIS FEATURE

DreamMaoMao:
As I mentioned in another pr, instead of expanding the number of tags, it is better to use a more flexible dynamic tagging system. However, this workload is extremely huge, and I still have other more important work to do

hboetes:
I completely understand the desire for a proper dynamic tagging system, that would be the right long-term solution. Since that work is still on the horizon though, would it make sense to merge this as an interim measure? It can always be refactored once the dynamic system lands, and in the meantime it unblocks current users.

Thanks for your consideration.

gplusplus314:
Hi @DreamMaoMao , I've updated this PR to be compatible with the latest `config.*` refactor that got rid of the global variables. This PR is small, doesn't re-architect anything, doesn't break anything, doesn't get in the way of a future refactor, and people can immediately benefit from it. Please reconsider your position.

If you're absolutely against this PR and will only ever accept a change with a more flexible dynamic tagging system, then it would be helpful to share what you had in mind in the form of a design document or discussion. Then we can see who'd be willing to work on it, how long it'll take, testing effort, etc. Until then, this PR exists right now and I'm happy to make any changes you'd like.

hboetes:
For anyone else wanting to keep up with HEAD and also wanting to use 10 workspaces, my original PR still applies and works:

https://github.com/mangowm/mango/pull/676.diff

gplusplus314:
Considering the state of the project and the need to preserve my own productivity, I’ve moved on from Mango, so I won’t be updating this branch. Someone else is more than welcome to resolve the merge conflicts. I don’t care about authorship credit - if you can get value out of my code, go ahead and take it.

DreamMaoMao:
done in latest commit.

jerrygreen:
Why force pushed instead of merging PR tho? By done you mean the work is done or you’re done and won’t do it? 😅

<s>How do I set more tags? With an option? Which option? I need like at least 25 tags I think.</s>

EDIT: apparently the option is called `tag_count` and allows anything to 32 tags, plus requires restart.

If it’s possible in Mango now, I might give it a second try :) as soon as it lands in Nixpkgs

DreamMaoMao:
refer this:

819371f85ac0fc940c3447ec14af62bd105af4cb

PR #756, opened 2026-03-11
Add mmsg command to get currently focused window's PID

grecodavide:
I think a way to get the currently focused window's PID would be useful (for example, if a window hangs it is useful to retreive the PID to kill it). This does just that by adding a new socket, which then just gets read


DreamMaoMao:
For the time being, I have no intention of changing this soon-to-be-deprecated protocol. Every change it makes will destroy all other dependent clients. I would rather directly add an action to force the termination of this window proces

grecodavide:
Ok, we can close this PR and I'll open another one with that implementation then.

PR #759, opened 2026-03-13
feat: add docs and sync with wiki & website

xtheeq:
Syncs the website docs and wiki from a single source.
Example wiki: https://github.com/atheeq-rhxn/mango/wiki

**Requires:** `WEBSITE_SYNC_TOKEN` repository secret
(Settings → Security → Secrets and variables → Actions)

The secret should be a **[Fine-grained Personal Access Token (PAT)](https://github.com/settings/personal-access-tokens)** with repository access to `mangowm/mangowm.github.io` and the following permission:

* **Contents:** Read and write

**Note:** The pr docs content is taken from the website documentation content.

### DIFF
--- a/.github/scripts/sync-wiki.py
+++ b/.github/scripts/sync-wiki.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+import json
+import re
+from pathlib import Path
+
+DOCS_DIR = Path("docs")
+WIKI_DIR = Path("wiki-temp")
+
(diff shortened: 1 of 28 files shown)


DreamMaoMao:
If the token is written in the source code, won't others be able to make any changes to the repository at will after getting this token?

xtheeq:
> If the token is written in the source code, won't others be able to make any changes to the repository at will after getting this token?

The token is not written in the source code.
The code only access the token through the secret variable `WEBSITE_SYNC_TOKEN` and this only can be used by the github action.

This is set in the settings.

PR #762, opened 2026-03-13
Add force kill option

grecodavide:
With this we can pass to killclient the "force" argument, to force kill the current window


BrockoliniMorgan:
This would be a great feature! Would you be able to add a description of the usage to the docs [here](https://github.com/mangowm/mango/blob/main/docs/bindings/keys.md?plain=1#L91)?

grecodavide:
Done :) thanks for pointing that out

PR #800, opened 2026-03-22
Add session restore support

cheerfulScumbag:
Adds opt-in session save/restore support to Mango.

  When `session_restore=1`:
  - Mango saves a restorable snapshot on clean shutdown
  - Mango loads that snapshot on next startup
  - Mango relaunches supported apps from persisted `launch_command`
  - Mango restores tag, monitor, floating/fullscreen state, and floating geometry as windows map

  When `session_restore=0`:
  - Mango behaves as before
  - no session file is written
  - no restore or relaunch path runs

  Supported config:
'session_restore=0|1'

  Optional override config:
'session_launch=app_id|command'
'session_launch=app_id|title|command'

  Current launch-command priority:
  1. exact command Mango itself used to spawn the app
  2. normalized higher-level launcher command recovered from process metadata
  3. raw process argv fallback
  4. title-specific `session_launch`
  5. app-only `session_launch`
  6. plain `app_id`

  Normalization currently implemented:
  - Flatpak apps normalize to desktop-entry launcher command
  - desktop-entry-backed apps normalize to the desktop file `Exec=` command

  Validated:
  - disabled path writes no session file
  - clean shutdown save
  - self-relaunch on next Mango startup
  - duplicate `foot` windows with title-specific mappings
  - exact command capture for Mango-owned `exec-once` launches
  - process-based recovery for apps Mango did not spawn directly
  - single-monitor restore
  - floating geometry restore
  - tag restore
  - two-monitor restore when outputs already exist before clients map

### DIFF
--- a/assets/config.conf
+++ b/assets/config.conf
@@ -92,10 +92,14 @@ idleinhibit_ignore_visible=0
 # Disabled by default to preserve Mango's current behavior.
 # Minimized restore is currently unsupported.
 # Restore expects target outputs to already exist before clients map.
+# Mango only restores from a trusted session file owned by the current user
+# and not writable by group or others.
 session_restore=0
-# Relaunch mapping used when Mango cannot infer a launch command from a client.
+# Optional relaunch override used when Mango cannot recover a suitable launcher.
 # Format: session_launch=app_id|command
 #    or: session_launch=app_id|title|command
+# Mango prefers exact Mango-owned spawn commands, then normalized desktop/Flatpak
+# launchers, then raw process argv before consulting session_launch.
 # session_launch=foot|foot
 # session_launch=foot|gamma|foot -a foot -T gamma -e sh -lc "sleep 600"
 # session_launch=org.kde.dolphin|dolphin .
--- a/assets/config.conf
+++ b/assets/config.conf
@@ -88,21 +88,11 @@ no_border_when_single=0
 axis_bind_apply_timeout=100
 focus_on_activate=1
 idleinhibit_ignore_visible=0
-# Restore and save session state across compositor restarts.
-# Disabled by default to preserve Mango's current behavior.
-# Minimized restore is currently unsupported.
-# Restore expects target outputs to already exist before clients map.
-# Mango only restores from a trusted session file owned by the current user
-# and not writable by group or others.
+# Opt-in session restore. See the Mango miscellaneous configuration docs.
 session_restore=0
-# Optional relaunch override used when Mango cannot recover a suitable launcher.
-# Format: session_launch=app_id|command
-#    or: session_launch=app_id|title|command
-# Mango prefers exact Mango-owned spawn commands, then normalized desktop/Flatpak
-# launchers, then raw process argv before consulting session_launch.
+# Optional overrides: session_launch=app_id|command
+#                 or: session_launch=app_id|title|command
 # session_launch=foot|foot
-# session_launch=foot|gamma|foot -a foot -T gamma -e sh -lc "sleep 600"
-# session_launch=org.kde.dolphin|dolphin .
 sloppyfocus=1
 warpcursor=1
 focus_cross_monitor=0
--- a/assets/config.conf
+++ b/assets/config.conf
@@ -88,6 +88,17 @@ no_border_when_single=0
 axis_bind_apply_timeout=100
 focus_on_activate=1
 idleinhibit_ignore_visible=0
+# Restore and save session state across compositor restarts.
+# Disabled by default to preserve Mango's current behavior.
+# Minimized restore is currently unsupported.
+# Restore expects target outputs to already exist before clients map.
+session_restore=0
+# Relaunch mapping used when Mango cannot infer a launch command from a client.
+# Format: session_launch=app_id|command
+#    or: session_launch=app_id|title|command
+# session_launch=foot|foot
+# session_launch=foot|gamma|foot -a foot -T gamma -e sh -lc "sleep 600"
+# session_launch=org.kde.dolphin|dolphin .
 sloppyfocus=1
 warpcursor=1
 focus_cross_monitor=0
--- a/assets/config.conf
+++ b/assets/config.conf
@@ -88,7 +88,6 @@ no_border_when_single=0
 axis_bind_apply_timeout=100
 focus_on_activate=1
 idleinhibit_ignore_visible=0
-# Opt-in session restore. See the Mango miscellaneous configuration docs.
 session_restore=0
 # Optional overrides: session_launch=app_id|command
 #                 or: session_launch=app_id|title|command
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -28,14 +28,9 @@ stdenv.mkDerivation {
   pname = "mango";
   version = "nightly";
 
-  src = lib.cleanSourceWith {
-    src = ../.;
-    filter =
-      path: type:
-      let
-        relPath = lib.removePrefix ((toString ../.) + "/") (toString path);
-      in
-      !(relPath == ".git" || lib.hasPrefix ".git/" relPath || relPath == "build" || lib.hasPrefix "build/" relPath);
+  src = builtins.path {
+    path = ../.;
+    name = "source";
   };
 
   mesonFlags = [
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -28,9 +28,14 @@ stdenv.mkDerivation {
   pname = "mango";
   version = "nightly";
 
-  src = builtins.path {
-    path = ../.;
-    name = "source";
+  src = lib.cleanSourceWith {
+    src = ../.;
+    filter =
+      path: type:
+      let
+        relPath = lib.removePrefix ((toString ../.) + "/") (toString path);
+      in
+      !(relPath == ".git" || lib.hasPrefix ".git/" relPath || relPath == "build" || lib.hasPrefix "build/" relPath);
   };
 
   mesonFlags = [
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -1796,10 +1796,10 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
 		size_t app_len, title_len = 0, cmd_len;
 
 		if (!first_sep) {
-			fprintf(stderr,
-					"\033[1m\033[31m[ERROR]:\033[33m Invalid session_launch "
-					"format. Expected app_id|command or "
-					"app_id|title|command\033[0m\n");
+			mango_error(
+				false, WLR_ERROR,
+				"Invalid session_launch format. Expected app_id|command or "
+				"app_id|title|command\n");
 			return false;
 		}
 
@@ -1815,9 +1815,8 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
 		if (app_len == 0 || cmd_len == 0 || app_len >= sizeof(rule.app_id) ||
 			cmd_len >= sizeof(rule.command) ||
 			title_len >= sizeof(rule.title)) {
-			fprintf(stderr,
-					"\033[1m\033[31m[ERROR]:\033[33m Invalid session_launch "
-					"entry length\033[0m\n");
+			mango_error(false, WLR_ERROR,
+						"Invalid session_launch entry length\n");
 			return false;
 		}
 
@@ -1834,9 +1833,8 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
 		trim_whitespace(rule.title);
 		trim_whitespace(rule.command);
 		if (rule.app_id[0] == '\0' || rule.command[0] == '\0') {
-			fprintf(stderr,
-					"\033[1m\033[31m[ERROR]:\033[33m session_launch requires "
-					"both app_id and command\033[0m\n");
+			mango_error(false, WLR_ERROR,
+						"session_launch requires both app_id and command\n");
 			return false;
 		}
 
@@ -1844,13 +1842,13 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
 							sizeof(*config->session_launch_rules) *
 								(config->session_launch_rules_count + 1));
 		if (!new_rules) {
-			fprintf(stderr,
-					"\033[1m\033[31m[ERROR]:\033[33m Failed to allocate "
-					"session_launch rules\033[0m\n");
+			mango_error(false, WLR_ERROR,
+						"Failed to allocate session_launch rules\n");
 			return false;
 		}
 		config->session_launch_rules = new_rules;
-		config->session_launch_rules[config->session_launch_rules_count++] = rule;
+		config->session_launch_rules[config->session_launch_rules_count++] =
+			rule;
 	} else if (strcmp(key, "no_border_when_single") == 0) {
 		config->no_border_when_single = atoi(value);
 	} else if (strcmp(key, "no_radius_when_single") == 0) {
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -62,6 +62,12 @@ typedef struct {
 } ConfigEnv;
 
 typedef struct {
+	char app_id[256];
+	char title[512];
+	char command[1024];
+} ConfigSessionLaunchRule;
+
+typedef struct {
 	const char *id;
 	const char *title;
 	uint32_t tags;
@@ -360,6 +366,7 @@ typedef struct {
 
 	uint32_t axis_bind_apply_timeout;
 	uint32_t focus_on_activate;
+	int32_t session_restore;
 	int32_t idleinhibit_ignore_visible;
 	int32_t sloppyfocus;
 	int32_t warpcursor;
@@ -492,6 +499,8 @@ typedef struct {
 
 	char **exec_once;
 	int32_t exec_once_count;
+	ConfigSessionLaunchRule *session_launch_rules;
+	int32_t session_launch_rules_count;
 
 	char *cursor_theme;
 	uint32_t cursor_size;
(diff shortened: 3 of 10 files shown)


cheerfulScumbag:
current implementation works, but I want to improve automatic launch recovery so common apps do not require as much manual session_launch configuration

cheerfulScumbag:
Updated this draft with a stronger automatic launch-command recovery path.

  Since the original draft, Mango now:
  - tracks exact commands for apps it launches itself
  - recovers commands from client process metadata for apps it did not launch directly
  - normalizes Flatpak apps to desktop-entry launcher command
  - normalizes desktop-entry-backed apps to the desktop file `Exec=` command

  That reduces the need for manual `session_launch` rules for common apps.

cheerfulScumbag:
tested on multiple programs and they all restore with no issues
so when `session_restore=1`:
  - Mango saves a restorable snapshot on clean shutdown
  - Mango loads that snapshot on next startup
  - Mango relaunches supported apps from persisted `launch_command`
  - Mango restores tag, monitor, floating/fullscreen state, and floating geometry as windows map
  
Give it a test and let me know what you think

cheerfulScumbag:
Small follow-up: I have been using this for a while now with no bugs i can find,  multi-monitor restore works fine.

 I also tightened the branch since the original draft with portable desktop-entry lookup, improved launcher recovery, and a trusted session-file check.

 @DreamMaoMao  whenever you have time, I’d appreciate any feedback on this.

DreamMaoMao:
I'm still not sure if I want to add this feature.

cheerfulScumbag:
Thanks for the reply.

Understood. To help me judge whether this is worth pressing  further,  can i ask what your hesitation mainly about?

I’m happy to reduce scope further if needed, but if the concern is that session restore itself is out of scope for Mango, that’s also useful for me to know.

src/mango.c
```diff
@@ -2291,6 +3246,8 @@ void cleanuplisteners(void) {
 }
 
 void cleanup(void) {
+	session_save_now(true);
+	session_shutdown();
```
Wateir, line 3661: Maybe i'm wrong on this, but from my understanding this two function gonna be call for everyone and not only the one who have session restore on the config

src/mango.c
```diff
@@ -555,6 +557,13 @@ typedef struct {
 	struct wl_listener destroy;
 } SessionLock;
 
+typedef struct {
+	struct wl_list link;
+	pid_t pid;
+	time_t created_at;
+	char command[1024];
```
Wateir: The 1024 here seems a magic number

src/session/session.h
```diff
@@ -0,0 +1,39 @@
...
+	int32_t width;
+	int32_t height;
+} SessionRect;
+
+typedef struct {
+	char app_id[256];
+	char title[512];
+	char monitor[128];
+	char launch_command[1024];
+	int32_t pid;
```
Wateir, line 26: Magic number here

src/mango.c
```diff
@@ -757,6 +766,17 @@ static void init_fadeout_layers(LayerSurface *l);
...
+void mango_session_spawn_tracker_shutdown(void);
+void mango_session_track_spawned_command(pid_t pid, const char *command);
+void mango_session_attach_spawn_command(Client *c);
+void mango_session_remember_client_launch_command(Client *c,
+												  const char *command);
+static char *mango_session_recover_process_command(pid_t pid);
+static void mango_session_attach_process_command(Client *c);
+static char *mango_session_find_desktop_exec(const char *app_id);
+static char *mango_session_find_flatpak_command(const char *app_id);
+static char *mango_session_normalize_launch_command(Client *c, pid_t pid);
```
Wateir, line 902: WHAT THE FUCK DO I WROTE ?

Sorry, idk why i'm thinking function declaration was global constant ....

src/mango.c
cheerfulScumbag: This is intentional. The session hooks are always called from the compositor lifecycle, but the feature gate is inside the session module.

src/mango.c
cheerfulScumbag: Good point. I’ll replace those with named constants so the limits are defined in one place instead of repeating raw buffer sizes.

dpvpro:
Very useful feature

src/mango.c
cheerfulScumbag: > mag

I replaced the repeated raw buffer sizes with named constants so the limits are defined in one place and reused across the session code.

cheerfulScumbag:
> Very useful feature

Thanks. I still think the feature is valuable, but I’m waiting on @DreamMaoMao  feedback about whether this fits Mango’s scope upstream.

cheerfulScumbag:
I’ve kept the branch rebased on current main, fixed the review comments that came up, and re-
  tested the Nix build after the recent upstream changes.

  I know you said you were unsure about adding session restore. Just wondering if this feels out
  of scope for Mango, or if there’s a smaller version / different approach that would be easier to
  consider.

assets/config.conf
```diff
@@ -88,6 +88,21 @@ no_border_when_single=0
...
+# and not writable by group or others.
+session_restore=0
+# Optional relaunch override used when Mango cannot recover a suitable launcher.
+# Format: session_launch=app_id|command
+#    or: session_launch=app_id|title|command
+# Mango prefers exact Mango-owned spawn commands, then normalized desktop/Flatpak
+# launchers, then raw process argv before consulting session_launch.
+# session_launch=foot|foot
+# session_launch=foot|gamma|foot -a foot -T gamma -e sh -lc "sleep 600"
+# session_launch=org.kde.dolphin|dolphin .
```
Wateir: This is quite weird, i belive it's more something who need to be on [mango doc](https://mangowm.github.io/docs) rather than here, since it's not here for other option

Btw why not use a multiline comment ?

docs/visuals/status-bar.md
```diff
@@ -28,8 +28,8 @@ Add the following to your Waybar configuration:
       "overview-label": "OVERVIEW",
   },
   "mango/keymode": {
-  	"format": "{}",
-  	// "format-default": " Default",
+    "format": "{}",
+    // "format-default": " Default",
```
Wateir: A reason to change tab size ? lintter configure differently ?

It's the only change on that file, maybe not need

docs/faq.md
```diff
@@ -65,7 +65,7 @@ Then force it for your specific game:
 windowrule=force_tearing:1,title:Counter-Strike 2
 ```
 
-> **Warning:** Some graphics cards require setting `env=WLR_DRM_NO_ATOMIC,1` in config before mango starts for tearing to work. 
+> **Warning:** Some graphics cards require setting `env=WLR_DRM_NO_ATOMIC,1` in config before mango starts for tearing to work.
```
Wateir: I don't think this need to be change, or one another PR, better to let the lintter work for another one

nix/default.nix
```diff
@@ -28,9 +28,14 @@ stdenv.mkDerivation {
...
-    path = ../.;
-    name = "source";
+  src = lib.cleanSourceWith {
+    src = ../.;
+    filter =
+      path: type:
+      let
+        relPath = lib.removePrefix ((toString ../.) + "/") (toString path);
+      in
+      !(relPath == ".git" || lib.hasPrefix ".git/" relPath || relPath == "build" || lib.hasPrefix "build/" relPath);
```
Wateir: Same, nix change need to be on other PR i belive.

I'm too bad at nix to really understand what going here, more something for @Ruixi-rebirth

src/config/parse_config.h
```diff
@@ -57,6 +57,12 @@ typedef struct {
 	char *value;
 } ConfigEnv;
 
+typedef struct {
+	char app_id[256];
+	char title[512];
+	char command[1024];
+} ConfigSessionLaunchRule;
```
Wateir, line 66: Not a big fan of magic number, but it's not something really special on this code base, so not really important

src/session/session.c
```diff
@@ -0,0 +1,519 @@
...
+typedef struct {
+	SessionRestoreEntry entry;
+	bool used;
+} PendingSessionEntry;
+
+static PendingSessionEntry *pending_entries;
+static size_t pending_count;
+static bool restore_started;
+
+static bool mkdir_p(const char *dir) {
```
Wateir: i see why you name this like that, it's because `mkdir -p` but i more explicti name who have no link to shell command is better i belive, something like `create_directory`, or maybe `create_session_restore_dir`

src/session/session.c
```diff
@@ -0,0 +1,519 @@
...
+	return true;
+}
+
+static void free_pending_entries(void) {
+	free(pending_entries);
+	pending_entries = NULL;
+	pending_count = 0;
+}
+
+static const char *skip_ws(const char *p) {
```
Wateir: Idk what do this function, i don't understand the name, seems something to skip to the next none special char

src/session/session.c
```diff
@@ -0,0 +1,519 @@
...
+	while (p && (*p == ' ' || *p == '\n' || *p == '\r' || *p == '\t'))
+		++p;
+	return p;
+}
+
+static const char *find_matching_brace(const char *start) {
```
Wateir: This function have more place on `parse_config.c` than here, maybe because nobody need it before you, but seems more logic to place on the parsing code function to parse code

src/session/session.c
```diff
@@ -0,0 +1,519 @@
...
+	p = strchr(p, ':');
+	if (!p)
+		return false;
+	p = skip_ws(p + 1);
+	if (!p || *p != '{')
+		return false;
+	*start = p;
+	*end = find_matching_brace(p);
+	return *end != NULL;
+}
```
Wateir: Mango use libcjson according to the [build dependency](https://github.com/mangowm/mango/blob/b5efdd24e76c3c9aa2300e626fe013fcd3617cca/meson.build#L40)

So i belive it's better to use the lib than to reinvent the wheel

src/session/session.c
```diff
@@ -0,0 +1,519 @@
...
+	if (!path)
+		return false;
+
+	if (access(path, F_OK) != 0)
+		goto cleanup;
+
+	if (!session_file_is_trusted(path)) {
+		fprintf(stderr,
+				"mango session: refusing to restore from untrusted session file: %s\n",
+				path);
```
Wateir: why it's not something like this ?
```C
fprintf(stderr,
					"\033[1m\033[31m[ERROR]:\033[33m mango session: refusing to restore from untrusted session file: %s\n",
				path);
```

src/mango.c
```diff
@@ -1199,6 +1219,944 @@ static struct wl_event_source *sync_keymap;
 #include "layout/scroll.h"
 #include "layout/vertical.h"
 
+static void session_write_json_string(FILE *out, const char *str) {
```
Wateir: Again, i belive it's better to use the libcjson lib that to reinvent the wheel

src/mango.c
```diff
@@ -1199,6 +1219,944 @@ static struct wl_event_source *sync_keymap;
...
+	}
+}
+
+void mango_session_spawn_tracker_init(void) {
+	wl_list_init(&session_spawn_commands);
+}
+
+void mango_session_spawn_tracker_shutdown(void) {
+	mango_session_spawn_tracker_cleanup(true);
+}
```
Wateir, line 1249: I belive it's bettter a macro here than a function, something like
```C
#define mango_session_spawn_tracker_init wl_list_init(&session_spawn_commands)
#define mango_session_spawn_tracker_shutdown mango_session_spawn_tracker_cleanup(true)
```

don't forget to undefine them to not be include on other file (i belive you don't want this behaviour, even if it's not static function, maybe i'm wrong)

src/mango.c
```diff
@@ -1199,6 +1219,944 @@ static struct wl_event_source *sync_keymap;
...
+	pid = fork();
+	if (pid == 0) {
+		signal(SIGSEGV, SIG_IGN);
+		signal(SIGABRT, SIG_IGN);
+		signal(SIGILL, SIG_IGN);
+
+		dup2(STDERR_FILENO, STDOUT_FILENO);
+		setsid();
+
+		execlp("sh", "sh", "-c", command, (char *)NULL);
```
Wateir, line 1965: `sh` is not a shell, you don't really know what is really behind, for most of people gonna be bash, but maybe weird people (like me) gonna have symlink to other thing, like dash, tsh or idk what, since you already inforce the need to have bash on the next line, inforce it for this one to to be sure of the behavior

src/mango.c
```diff
@@ -1199,6 +1219,944 @@ static struct wl_event_source *sync_keymap;
...
+		session_write_json_string(out, appid);
+		fputs(",\n    \"title\": ", out);
+		session_write_json_string(out, title);
+		fprintf(out,
+				",\n    \"pid\": %d,\n    \"monitor\": ",
+				client_get_pid(c));
+		session_write_json_string(out, monitor_name);
+		fputs(",\n    \"launch_command\": ", out);
+		session_write_json_string(out, launch_command);
+		fprintf(out,
```
Wateir: I'm pretty sure it's exist a better way to do it than that

Wateir (commented):
Hey, so here we are. I'm sorry i don't really look into the code deeply, just a general overview, hope it's what you wanted.

The code seems pretty clear, just some little thing who bother me, but maybe because school is hard with me on that kind of thing.

TLDR: use the lib cjson, some function name not really explicit and little quirk not major here and here

assets/config.conf
cheerfulScumbag: Good point. I shortened the sample configuration and moved the detailed behavior and limitations into the miscellaneous configuration documentation. I retained `#` comments because that is Mango configuration syntax.

src/config/parse_config.h
cheerfulScumbag: I consolidated the session limits into shared named constants for app IDs, titles, monitor names, and commands, and reused them in both persisted entries and `session_launch` rules.

docs/visuals/status-bar.md
cheerfulScumbag: Agreed. This was unrelated whitespace noise, so I reverted it.

docs/faq.md
cheerfulScumbag: Agreed. I reverted this unrelated change.

src/session/session.c
cheerfulScumbag: Agreed. I renamed it to `ensure_directory_tree`, which describes the behavior without referring to a shell command.

src/session/session.c
cheerfulScumbag: Agreed. The handwritten JSON parser has been replaced with libcjson, including type, range, required-field, and buffer-length validation.

src/session/session.c
cheerfulScumbag: This helper has been removed. Session-file parsing now uses libcjson.

src/mango.c
cheerfulScumbag: I kept these as functions intentionally. They cross the session-module boundary, provide type checking and debugger-visible symbols, and avoid exposing implementation details through macros.

src/mango.c
cheerfulScumbag: I kept `sh` deliberately. It is the portable POSIX shell interface and Mango already uses the same execution pattern elsewhere. Requiring Bash would introduce an unnecessary portability dependency.

src/session/session.c
cheerfulScumbag: I changed this to `wlr_log(WLR_ERROR, ...)`. That follows the Mango runtime logging path and avoids embedding terminal colour escapes in redirected logs.

nix/default.nix
cheerfulScumbag: Agreed. I removed the Nix change from this PR so it can be considered separately if it is still needed.

src/mango.c
cheerfulScumbag: Agreed. Snapshot generation now constructs the complete JSON array and entries with libcjson, and serialization failures cannot replace the previous session file.

src/mango.c
cheerfulScumbag: Agreed. Manual JSON string escaping has been removed; libcjson now handles serialization.

src/session/session.c
cheerfulScumbag: The brace parser has been removed as part of the libcjson conversion. Session JSON remains in the session module because it parses persisted session state rather than Mango configuration.

cheerfulScumbag:
Thanks for taking the time to review this. In a96e8b2 I switched session persistence to the existing libcjson dependency, removed the unrelated documentation and Nix changes, moved the detailed configuration explanation into the documentation, consolidated the buffer limits, improved naming, and changed runtime errors to use wlr_log.

The updated code builds successfully and passed isolated live tests for disabled mode, clean-shutdown save, automatic relaunch, quoted and Unicode values, placement matching, invalid entries, and untrusted session files.

src/mango.c
Wateir: Fair

src/mango.c
Wateir: But https://github.com/mangowm/mango/pull/800/changes/BASE..a2d59c15cd18b943a41b6b933ebfc4b3365fd6be#diff-84f9cb64166644d880e9c19de9fd16e20e7b5983c10d6b9f609130f0411aea1cR1996 here (litteraly the line after), you enforce bash, that why i suggest this

cheerfulScumbag: The Bash call is only a fallback: a successful `execlp("sh", ...)` never returns, so it is reached only if executing `sh` fails. Commands are normally interpreted by the system POSIX `sh`; Bash is not required or enforced. This also intentionally matches Mango's existing `spawn_with_shell` implementation on `main`.

cheerfulScumbag:
I opened a [community poll](https://github.com/mangowm/mango/discussions/1218) to find out whether Mango users want session restore and to discuss whether the existing implementation or `xdg-session-management-v1` would be the better upstream direction. I’ll keep this PR open while that discussion happens.

cheerfulScumbag:
hey @Wateir friendly bump on this when you have a moment.

assets/config.conf
```diff
@@ -89,6 +89,11 @@ no_border_when_single=0
 axis_bind_apply_timeout=100
 focus_on_activate=1
 idleinhibit_ignore_visible=0
+# Opt-in session restore. See the Mango miscellaneous configuration docs.
```
Wateir: This comment are quite unique, for consitency with other maybe remove it

src/config/parse_config.h
```diff
@@ -1728,6 +1739,71 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
...
+	} else if (strcmp(key, "session_restore") == 0) {
+		config->session_restore = atoi(value);
+	} else if (strcmp(key, "session_launch") == 0) {
+		ConfigSessionLaunchRule rule = {0};
+		ConfigSessionLaunchRule *new_rules = NULL;
+		char *first_sep = strchr(value, '|');
+		char *second_sep = NULL;
+		size_t app_len, title_len = 0, cmd_len;
+
+		if (!first_sep) {
```
Wateir, line 1798: sorry, but my pr to add "mango_error" have been merge, can you use it now ?

Wateir (changes requested):
I'm sorry i can't read longer i miss time.

But i'm really interest on you're pr, i'm gonna try to merge it on my hardfork of mango to test it and see.

Sorry to not have time to help you more.

Gonna come back on it when i have more time

cheerfulScumbag:
Thanks for taking a look and for testing it out on your fork!
 
 I've updated the PR with the requested changes:

• Rebased cleanly onto the latest main.
• Removed the extra comment line in assets/config.conf.
• Switched session_launch config error reporting to mango_error(...).

Take your time testing, and let me know if any questions or edge cases come up!

PR #810, opened 2026-03-26
fix(xwayland): improve pointer behavior for fullscreen games

### LINKED ISSUE #807: Problems with fullscreen games
mango version:mango 0.12.7(91e10b2)

Have a couple of issues preventing me from daily driving 

Fullscreen games lose exclusive fullscreen when switching windows via scroller (same tag)

Environment

- Mango Version: mango 0.12.7(91e10b2)
- Distro: CachyOS
- GPU: RX 6700xt
- Games Tested: Deadlock, Marvel Rivals

When running games in exclusive fullscreen, switching focus to another window using the scroller within the same tag causes the game to exit fullscreen (likely falling back to borderless or incorrect resolution state).

Some games fail to properly reinitialize their resolution after focus is restored, leading to rendering issues.

This issue doesnt happen when switching to another tag in scroller while inside the game, only when switching to another window inside a tag

Steps to Reproduce

1. Launch a game in exclusive fullscreen (e.g., Deadlock or Marvel Rivals)
2. Ensure another window exists in the same tag
3. Use the scroller to switch to that window
4. Switch back to the game

Expected Behavior

- Game remains in exclusive fullscreen,or If focus is lost, it should correctly restore fullscreen and resolution on return

Actual Behavior

- Game exits exclusive fullscreen when switching away, and goes back to default scroller layout size

On return:
Deadlock: resolution becomes incorrect / scaled improperly
Marvel Rivals: Rescales screen to fit default scroller layout, but not very satisfying to have to fullscreen again manually, should just stay in fullscreen like other apps

This issue doesnt happen when switching between tags, the games retain their fullscreen state

EduardoRodaRoca:
Description
This PR improves fullscreen behavior for Xwayland games on scaled outputs.
It adjusts pointer coordinates using the output scale before sending pointer enter and motion events to X11 clients.

Why
On HiDPI or fractional scaling, some fullscreen Xwayland games can receive incorrect mouse coordinates, which may cause offset or inconsistent cursor behavior.

Scope

- Applies only to Xwayland clients
- No change for native Wayland clients
- No effect when scale is 1.0

Mostly band-aid, as for some reason some games are not being treated as scale 1.0 even when in togglefullscreen, but this does fix some weird mouse behaviour with those games

Tested:

Marvel Rivals Launcher: Fixes mouse coordinates

Where Winds Meet launcher: Fixes Mouse coordinates

Deadlock: Fixes mouse coordinates in general for Menus

Fixes #807

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -4565,6 +4565,16 @@ void pointerfocus(Client *c, struct wlr_surface *surface, double sx, double sy,
 		time = now.tv_sec * 1000 + now.tv_nsec / 1000000;
 	}
 
+#ifdef XWAYLAND
+	if (c && c->mon && client_is_x11(c)) {
+		float scale = c->mon->wlr_output->scale;
+		if (scale > 0.0f && fabsf(scale - 1.0f) > 0.001f) {
+			sx *= scale;
+			sy *= scale;
+		}
+	}
+#endif
+
 	/* Let the client know that the mouse cursor has entered one
 	 * of its surfaces, and make keyboard focus follow if desired.
 	 * wlroots makes this a no-op if surface is already focused */


DreamMaoMao:
fix in 

https://github.com/mangowm/mango/commit/f94ddc671e1d01ad85ffd6f1cf2f09811c2629c8

PR #813, opened 2026-03-28
feat: add canvas layout

ernestoCruz05:
Adds a 2D infinite canvas-like layout mode that allows windows to be freely positioned and sized on a virutal canvas, with support for:
- Zooming in/out
- Minimap
- Overview of the canvas
- Touchpad navigation

Video showcase: https://streamable.com/20e5or
(Can't attach it to the message because it's too big)

Note: I understand this may not fit maomao's direction and if he believes it shouldn't be on the repo i'm happy to maintain it in my fork. From my testing, there were no issues in both AMD&NVIDIA, also I didn't see a big performance cost either


ernestoCruz05:
There are some issues with multi-monitor setups that i haven't been able to test for now, i'll fix them and re-open a new PR when all issues are fixed

PR #818, opened 2026-03-29
Additional canvas layout

ernestoCruz05:
Adds a 2D infinite canvas-like layout mode that allows windows to be freely positioned and sized on a virutal canvas, with support for:

    Zooming in/out
    Minimap
    Overview of the canvas
    Touchpad navigation

Video showcase: https://streamable.com/20e5or
(Can't attach it to the message because it's too big)

Note: I understand this may not fit maomao's direction and if he believes it shouldn't be on the repo i'm happy to maintain it in my fork. From my testing, there were no issues in both AMD&NVIDIA, also I didn't see a big performance cost either.

Also big thanks to June and wAlt from the mangowm discord for helping me test it out and finding weird bugs that i couldn't find!


DreamMaoMao:
To be honest, I'm not very inclined to merge this one because mango is mainly keyboard-driven and it needs to introduce a lot of special logic for compatibility. 

By the way, I roughly took a look at the code and found that this scaling method actually has problems Because it is equivalent to requiring the gpu to forcibly scale the buffer on the client surface once in each frame, this will put a heavy load on the gpu,you can check the gpu usage rate during scaling for comparison.

ernestoCruz05:
Yeah i had the feeling you would feel this way (completely understandable ofc), a lot of edge cases had to be handled to not cause any issues and i feel like in future development the canvas might also need to be taken in mind and that's more chores for the awesome work you do... 

Regarding the keyboard-driven aspect, i'm working on a "tiling" solution that would make windows spawn close to each other both vertically and horizontally creating something like a grid and the user can just use focusdir to move around in the canvas, people who like using their mouse can sitll pan around.

Finally on the GPU usage, seems like I was testing wrongly before and wasn't seeing the iGPU usage(about 60% spike when zooming out on my machine), if the user is just zoomed out and moves the viewport i didn't any performance hits, maybe i'm testing it wrong again,..still on lower end computers i assume it's even worse, idk if there is a way to make this better sadly

ernestoCruz05:
Since we talked about keyboard-centric movement, i think this might be a good fix for that...
https://github.com/user-attachments/assets/afcbe5b8-1096-420c-814e-cb2fcc56a1b1

PR #821, opened 2026-03-31
nix: bump scenefx

jibi:
there's currently an evaluation warning:

    evaluation warning: The xorg package set has been deprecated, 'xorg.libxcb' has been renamed to 'libxcb'
    evaluation warning: The xorg package set has been deprecated, 'xorg.xcbutilwm' has been renamed to 'libxcb-wm'

commit wlrfx/scenefx@6291016 updated some package names in line with upstream changes, and bumping that flake will make the warning disappear.

Unfortunately this means we need to deal with a couple of breaking APIs changes:

* the `enum corner_location` and `int corner_radius` pair has been replaced with `struct fx_corner_radii`
* `wlr_scene_buffer_set_backdrop_blur()`, `_ignore_transparent()`, and `_optimized()` have been removed, and  blur is now managed at the scene level via `wlr_scene_blur` nodes

moreover, it appears that https://github.com/wlrfx/scenefx/commit/51d9081 introduced a regression with how transparency is handled, so bump scenefx only to the commit before that (https://github.com/wlrfx/scenefx/commit/bdc3e64) rather than main.

Note that I asked claude for some help with these changes, as I have zero experience with scenefx or wayland. While the first set of changes (`struct fx_corner_radii`) looks reasonable to me, the rest (`wls_scene_blur_*`) probably needs a more thorough look

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,6 +18,27 @@
         "type": "github"
       }
     },
+    "flake-utils": {
+      "inputs": {
+        "systems": [
+          "scenefx",
+          "systems"
+        ]
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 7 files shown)


Yappaholic:
Seems like it's not that much related to nix and more about bumping scenefx version.

Currently the scenefx target is 0.4.1 since it is widely distributed and new versions introduce some performance regressions, so I'd think that waiting for next stable version is a way to go

jibi:
> Seems like it's not that much related to nix and more about bumping scenefx version.
> 
> Currently the scenefx target is 0.4.1 since it is widely distributed and new versions introduce some performance regressions, so I'd think that waiting for next stable version is a way to go

Ack, makes sense :+1:  the initial intent was to suppress those warnings, but the bump ended up requiring a bunch of extra scenefx changes

PR #826, opened 2026-04-04
Multiple Guix package fixes

ch4og:
This series of commits aligns the package closer with the official Guix package which was made by me.

* Pins the wlroots version. Guix will soon move from wlroots-0.19 to wlroots-0.20.
* Updates the `home-page` field to use the current upstream repository URL.
* Lists all bundled licenses.
* Replaces meson patching logic with a much easier one to read and maintain. Reuses variable for NixOS.
* Changes description to provide more context.


Yappaholic:
Thank you for the update, looks good to me

ch4og:
Force pushed update description commit to also move description string to newline which is Guix default style for description field.

PR #827, opened 2026-04-05
NixOS: wl-only branch bump wlroots_0_19 to wlroots_0_20

DrVoidest:
Good afternoon, all,
I just did a simple fix to the wl-only branch to bump the wlroots library from 0.19->0.20 so it builds again.
Edit: I tested to make sure it would build on my system first, and it builds.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -16,7 +16,7 @@
   meson,
   ninja,
   scenefx,
-  wlroots_0_19,
+  wlroots_0_20,
   libGL,
   enableXWayland ? true,
   debug ? false,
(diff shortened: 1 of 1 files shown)


dl09r:
But the wl-only branch always used 0.20. The main branch is still on 0.19 until scenefx is updated.

DrVoidest:
Right, I am only editing the wl-only branch default.nix. Which still used wlroots 0.19.

dl09r:
> Right, I am only editing the wl-only branch default.nix. Which still used wlroots 0.19.

Ah got it. Nix.

DreamMaoMao:
I'm waiting for the update of scenefx. At that time, both branches will be synchronized simultaneously

DrVoidest:
Ok, cool, thanks.

PR #836, opened 2026-04-08
docs: update recommended package name

therylith:
The upstream package 'swww' has been renamed to 'awww'.This commit:
- Updates the package name for rolling-release distributions like archlinux
- Keeps 'swww' for other distros that haven't updated yet.


Yappaholic (approved):
LGTM

PR #837, opened 2026-04-08
refactor(config parsing): changed from if/elif chain to precalculated hash table with a switch case and added caching of keysyms

DuckTapeMan35:
this refactor marginally speeds up config parsing by targetting the 2 biggest bottlenecks, namely iterating over keysyms every time and using extremely long if/elif chains.

benchmarking each command with 1000 runs:

/home/duck/mango/build/mango -p (the code in this PR)
  Average: 4.88 ms  (std dev: 1.08 ms)

mango -p
  Average: 5.31 ms  (std dev: 1.01 ms)

about 1.09x faster

this PR does somewhat make it harder to add new functions/options in that they need to be added to both parse_config.h in the switch case and in the appropriate .gperf file.

that being said the improvements would get more noticeable as mango adds more options to its configuration.

as for readability, I believe a switch case is more readable.


DreamMaoMao:
So is this just a few milliseconds faster? I'm not quite sure if it's worth it, because it introduces new dependencies.

DuckTapeMan35:
Yes. I wasn't sure if I should make a PR because it's not a very big improvement, but I did anyway to let you decide

PR #840, opened 2026-04-10
feat: add OpenBSD support

VlkrS:
OpenBSD does not have wordexp, therefore we need to add a reasonable approximation of mango's standard behavior when spawning commands.

Thank you in advance for considering it.

### DIFF
--- a/src/action/monitor.h
+++ b/src/action/monitor.h
@@ -13,6 +13,11 @@ bool mango_scene_output_commit(struct wlr_scene_output *scene_output,
 	if (!wlr_scene_output_build_state(scene_output, state, NULL))
 		return false;
 
+	if (m->need_init_hdr) {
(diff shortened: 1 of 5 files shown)


Yappaholic:
Hi, can this be added as a patch to OpenBSD ports (like FreeBSD does)?

VlkrS:
> Hi, can this be added as a patch to OpenBSD ports (like FreeBSD does)?

It's what we currently do: 
see https://cvsweb.openbsd.org/ports/wayland/mango/patches?sort=File

Some projects are interested in including such patches in their code, and therefore I am proposing it here. If for whatever reason, mango does not wish to do that, that is obviously a perfectly valid decision, and I will stop proposing them.

VlkrS:
Rebased on 0.14.0

VlkrS:
Rebased on 0.15.1

PR #842, opened 2026-04-11
refactor(parse_config.h): replace the if else chain with a LUT loop approach

ernestoCruz05:
Replaces the  giant if-else strcmp chain that made searching tedious

There should be no functional change, every command maps to the same function pointer and argument-parsing logic as before. Special-case commands like setoption, view, spawn, etc... each get a dedicated ARG_SPECIAL_* variant so their unique logic stays isolated in its own case block.

Adding now is just adding a new row to cmd_table[], If the arg shape is new, add an ArgParseType variant and a case in the switch, otherwise reuse an existing type.

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -920,11 +920,108 @@ uint32_t parse_num_type(char *str) {
 	}
 }
 
+typedef enum {
+	ARG_NONE,
+	ARG_INT,
+	ARG_FLOAT,
+	ARG_UINT,
+	ARG_DIRECTION,
+	ARG_CIRCLE_DIR,
+	ARG_UINT_BIT,
+	ARG_STR_V,
+	ARG_MOUSE_ACTION,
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
What are the benefits of doing this? Although the previous form was a bit unsightly, it is still very clear. Each time a action is added, only three lines of code need to be inserted in the if-else statement.

ernestoCruz05:
The main structural benefit is that argument-parsing logic is centralized in the switch, if the way a category of args is handled is for some reason changed it's one edit instead of finding every matching branch.

The benefit here is about future maintenance rather than readability today. Happy to close this if you'd rather keep  the original style.

ernestoCruz05:
@DreamMaoMao  if you don’t say anything else i will assume you don’t like the refactor and i will close the PR to not pollute the repo

DreamMaoMao:
sorry ,I'm busy with other work.

PR #848, opened 2026-04-13
fix : ensure the current dekstop is mango before executing ipc dispatch

Wateir:
This commit fix a issue that if you run `mmsg [flag] [param]` with a valide command on a compositor or a wm which is not mango but have mango install mmsg segfault.

Check if `XDG_CURRENT_DEKSTOP` is mango, then execute it, if not return a error and close the program

Test under Artix Linux x86_64

### DIFF
--- a/mmsg/mmsg.c
+++ b/mmsg/mmsg.c
@@ -546,6 +546,15 @@ static void usage(void) {
 }
 
 int32_t main(int32_t argc, char *argv[]) {
+
+	const char *env_XDG_DEKSTOP = getenv("XDG_CURRENT_DESKTOP");
+
+	if (!env_XDG_DEKSTOP || strcmp(env_XDG_DEKSTOP, "mango")) {
+		fprintf(stderr, "wrong dekstop, wanted 'mango' have '%s'\n",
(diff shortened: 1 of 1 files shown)


Matercan:
Some peoples' XDG_CURRENT_DESKTOP env var may be wlroots instead of mango.

Wateir:
> Some peoples' XDG_CURRENT_DESKTOP env var may be wlroots instead of mango.

Yeah but on this case, a bunch of compositor can be name wlroots, but i'm not against add other form of check than the xdg env var, i don't see any other simple right know

https://github.com/hyprwm/Hyprland/blob/933a24caa68d4d217207838bceabd5577838431c/hyprctl/src/main.cpp#L485-L491
for exemple hyprland use a env var specific `HYPRLAND_INSTANCE_SIGNATURE`, maybe something like that and use it after to check if mango is running

Wateir:
This allow user to have more complexe string than just `'mango'` but `'mango,wlroots'` or any other fantasy

Wateir:
https://github.com/mangowm/mango/pull/967 fix it

PR #858, opened 2026-04-18
Added silent variations of tagtoleft and tagtoright

lilly-lizard:
My preferred workflow is to bind `ALT+n` to `viewtoright` and `ALT+SHIFT+n` to `tagtorightsilent` so I can do one or the other or both easily.

I've made a private `tagtoleft_general` function to avoid code duplication and help maintainability in the future.

Note that I also tried to implement a silent version of `tagmon` but encountered some weird behavior around focusing that I've yet to figure out...


src/dispatch/bind_define.h
```diff
@@ -1193,30 +1193,66 @@ int32_t tagsilent(const Arg *arg) {
...
-int32_t tagtoright(const Arg *arg) {
+int32_t tagtoleft(const Arg *arg) {
+	return tagtoleft_general(arg, false);
+}
+
+int32_t tagtoleftsilent(const Arg *arg) {
+	return tagtoleft_general(arg, true);
+}
+
+int32_t tagtoright_general(const Arg *arg, bool silent) {
```
Wateir: A bunch of the code of this function is basicly the same as the one for 
`tagtoleft_general`, is not just possible to make a function name idk `tagtoside_general` and just pass it `target >>= 1;` on one case, and `target <<= 1;` on the other case ?

Wateir: Something like
```C
int32_t tagtoside_general(const Arg *arg, bool silent, bool left){
    if (!selmon)
		return 0;

	if (selmon->sel != NULL &&
		__builtin_popcount(selmon->tagset[selmon->seltags] & TAGMASK) == 1) {

		uint32_t target = selmon->tagset[selmon->seltags];
		if (target & 1) {
			// 1 wraps around to 9
			target |= 1 << LENGTH(tags);
		}
		if (left)
		    target >>= 1;
		else
		    target <<= 1;

		if (silent)
			tagsilent(&(Arg){.ui = target, .i = arg->i});
		else
			tag(&(Arg){.ui = target, .i = arg->i});
	}
	return 0;
   }
}

int32_t tagtoleft(const Arg *arg) {
	return tagtoside_general(arg, false, true);
}

int32_t tagtoleftsilent(const Arg *arg) {
	return tagtoside_general(arg, true, true);
}

int32_t tagtoright(const Arg *arg) {
	return tagtoside_general(arg, false, false);
}

int32_t tagtorightsilent(const Arg *arg) {
	return tagtoside_general(arg, true, false);
}
```
Continue you're idea of refractor code

lilly-lizard: good idea, have pushed a commit now using `tagtoside_general`. have tested and still works

PR #865, opened 2026-04-22
pointer constraint handling for layer-shell surfaces (lan-mouse input retain issue)

mdouda:
#### Purpose:
I am a heavy user of input capture software like Synergy, deskflow and [lan-mouse](https://github.com/feschber/lan-mouse). I'm in love with what mango offers but there was this little detail not letting me be 100% happy. 
lan-mouse worked partially, entering the second PC's screen properly. But when moving the mouse back, mango didn't wait for input release from lan-mouse and immediately released the input, moving the cursor back to the main PC's screen.

#### Problem:
lan-mouse's capture backend uses a layer-shell surface, not a toplevel window. toplevel_from_wlr_surface() returns NULL for layer-shell surfaces. This is because "if (c && ...)" block is always NULL.

#### Solution:
Reordering the checks in the block this way fixes the issue that lan-mouse has, releasing the input on an actual focus loss, not on loss of border friction.

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -4384,20 +4384,23 @@ void motionnotify(uint32_t time, struct wlr_input_device *device, double dx,
 
 		if (active_constraint && cursor_mode != CurResize &&
 			cursor_mode != CurMove) {
-			toplevel_from_wlr_surface(active_constraint->surface, &c, NULL);
-			if (c && active_constraint->surface ==
-						 seat->pointer_state.focused_surface) {
-				sx = cursor->x - c->geom.x - c->bw;
-				sy = cursor->y - c->geom.y - c->bw;
-				if (wlr_region_confine(&active_constraint->region, sx, sy,
-									   sx + dx, sy + dy, &sx_confined,
-									   &sy_confined)) {
-					dx = sx_confined - sx;
-					dy = sy_confined - sy;
-				}
+			if (active_constraint->surface ==
+				seat->pointer_state.focused_surface) {
 
 				if (active_constraint->type == WLR_POINTER_CONSTRAINT_V1_LOCKED)
 					return;
+
+				toplevel_from_wlr_surface(active_constraint->surface, &c, NULL);
+				if (c) {
+					sx = cursor->x - c->geom.x - c->bw;
+					sy = cursor->y - c->geom.y - c->bw;
+					if (wlr_region_confine(&active_constraint->region, sx, sy,
+										   sx + dx, sy + dy, &sx_confined,
+										   &sy_confined)) {
+						dx = sx_confined - sx;
+						dy = sy_confined - sy;
+					}
+				}
 			}
 		}
 


DreamMaoMao:
I didn't understand what you said
Moreover, from the result, your modification shows that the logic is completely equivalent and no functional or behavioral changes have been made

mdouda:
Sorry for the lacking explanation.
The whole idea is the base functionality of mango remains unchanged and there is 0 impact. This is good.
What this PR fixes is support for lan-mouse, which allows sharing the mouse and keyboard with other computers. With lan-mouse, you basically treat another computer almost as an additional screen, moving your mouse over seamlessly between them. The issue was that these checks, ordered as they were, did not allow for retention of the mouse on the second computer, releasing the input before lan-mouse decided it was time to do so.
I could add a short video demonstration to better show what the problem and the fix look like, if you think this would help :)

DreamMaoMao:
I didn't see any changes between the pr and the before code. It still went through the same judgment and performed the same actions in the end

mdouda:
Yes, but by just decoupling "if (c && active_constraint->surface == ..." into 2 separate statements fixed the input grab with lan-mouse. I guess order mattered there.
I could also keep this as a personal patch as well if you're not comfortable merging this. I just thought this may help someone with the same issues as me.

DreamMaoMao:
This doesn't make sense. Why can doing this solve the problem? In terms of the result, it doesn't change anything.

How did you know to change the order like this? Could you explain it to me? Maybe I missed something?

mdouda:
toplevel_from_wlr_surface() returns NULL for layer-shell surfaces as it's not toplevel. So c in this scenario is NULL, the whole if block is skipped including the WLR_POINTER_CONSTRAINT_V1_LOCKED. The lock constraint never fires, input is released and lan-mouse capture window is broken this way.
When checking surface first and separating the if(c...), LOCKED is reachable without the toplevel being required. 
This makes it work even if there's no xdg toplevel.

mdouda:
As per how:
I realized that lan-mouse partially works as is. The screen border is detected properly and capture starts.
Then I started digging into the mechanism that lan-mouse uses to make this happen. On stuff like plasma and gnome. I found it in their repo:
https://github.com/feschber/lan-mouse/blob/a878c985f0633a12b00a363883fd56385c2368e7/input-capture/src/layer_shell.rs#L73
Then I looked for what this means and found this resource for wl:
https://wayland.app/protocols/pointer-constraints-unstable-v1
And then I searched for something similar in mango.c and found the constraint.
It was only then when I began to understand the whole surface thing and ran into the fact that WLR_POINTER_CONSTRAINT_V1_LOCKED wouldn't trigger under its current conditions.

I'm no pro in wayland or window managers, so please excuse me if this is not the right way. I was merely trying to make this work and found this out.

DreamMaoMao:
well,I missed the below code last time.

PR #870, opened 2026-04-25
fix: keep XWayland frame callbacks alive on hidden workspaces

### LINKED ISSUE #867: xwayland FPS going to 0 when not focused
## Info
xwayland apps (some games) goes to 0 fps when the user focus on another workspace or another window (and the said xwayland app is not visible).

Steps to reproduce : 

1. Open a xwayland app
2. Change workspace
3. Notice the drop of FPS when going back to it

FWIW : There is something similar with hyprland and it can get fixed with the render_unfocused option

render_unfocused | [on] | Forces the window to think it’s being rendered when it’s not visible. See also Variables - Misc for setting render_unfocused_fps.
-- | -- | --

mango version: mango 0.12.8(4b86031)
wlroots version:

tayacrystals:
## Summary

Fixes #867 — XWayland apps (games) dropping to 0 FPS when switched to a hidden workspace.

- Instead of disabling the scene node (which stops frame callbacks), move XWayland clients to a far-offscreen position (`-100000, -100000`) while keeping the node enabled
- Applies at all hide sites: `clip_to_hide`, `client_animation_next_tick`, `set_arrange_hidden`
- Adds `xwayland_render_unfocused` config option (default `1`) to allow opting out
- Mirrors the approach used by Hyprland's `render_unfocused` option (referenced in the issue)

## Test plan

- [ ] Open an XWayland game or FPS-sensitive app
- [ ] Switch to another workspace
- [ ] Switch back — confirm FPS is maintained while on the other workspace
- [ ] Set `xwayland_render_unfocused = 0` in config and confirm old behavior is restored

### DIFF
--- a/src/animation/client.h
+++ b/src/animation/client.h
@@ -8,6 +8,29 @@ void set_rect_size(struct wlr_scene_rect *rect, int32_t width, int32_t height) {
 	wlr_scene_rect_set_size(rect, GEZERO(width), GEZERO(height));
 }
 
+/* Keep node enabled offscreen so XWayland frame callbacks keep flowing,
+ * preventing games from stalling at 0 FPS when on a hidden workspace. */
+#define XWAYLAND_OFFSCREEN_OFFSET (-100000)
+
+void xwayland_hide_offscreen(Client *c) {
+	if (!client_is_x11(c))
+		return;
+	c->is_xwayland_hidden = true;
+	wlr_scene_node_set_enabled(&c->border->node, false);
+	wlr_scene_node_set_enabled(&c->shadow->node, false);
+	wlr_scene_node_set_position(&c->scene->node, XWAYLAND_OFFSCREEN_OFFSET,
+								XWAYLAND_OFFSCREEN_OFFSET);
+}
+
+void xwayland_show_from_offscreen(Client *c) {
+	if (!client_is_x11(c) || !c->is_xwayland_hidden)
+		return;
+	c->is_xwayland_hidden = false;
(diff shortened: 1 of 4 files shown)


vinthara:
Hi taya,

Thanks for your work,
I have been trying to build and run your test scenario.
The FPS doesn't seem to render correctly either:

Here is the steps I took : 

```
git clone https://github.com/mangowm/mango.git ~/mango-test  
cd ~/mango-test
git fetch origin pull/870/head:pr-870
meson setup build
~/mango-test/build/mango
```

The config file updated : 
```
grep -r "xwayland_render_unfocused" ~/.config/mango/config.conf 
xwayland_render_unfocused=1
```

src/animation/client.h
```diff
@@ -8,6 +8,29 @@ void set_rect_size(struct wlr_scene_rect *rect, int32_t width, int32_t height) {
 	wlr_scene_rect_set_size(rect, GEZERO(width), GEZERO(height));
 }
 
+/* Keep node enabled offscreen so XWayland frame callbacks keep flowing,
+ * preventing games from stalling at 0 FPS when on a hidden workspace. */
+#define XWAYLAND_OFFSCREEN_OFFSET (-100000)
+
+void xwayland_hide_offscreen(Client *c) {
+	if (!client_is_x11(c))
+		return;
```
Wateir, line 17: Maybe change the signature of the function to return a int rather than a void and allow the calling function if need to check return code then.

DreamMaoMao:
This approach is quite strange. It disrupts some judgment logics, and it seems to be a bug on the client side.

PR #880, opened 2026-04-30
fix: caps lock problem resolver

dpvpro:
With this code i resolved issue with changing keyboard layout by Caps Lock key over xkb settings.
It seems that that code does not affected other things.

Related issues:
https://github.com/mangowm/mango/issues/868
https://github.com/noctalia-dev/noctalia-shell/issues/1639
https://github.com/noctalia-dev/noctalia-shell/issues/2572

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -460,6 +460,7 @@ typedef struct {
 	struct wl_listener modifiers;
 	struct wl_listener key;
 	struct wl_listener destroy;
+	xkb_layout_index_t last_group;
 } KeyboardGroup;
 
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
It seems very costly to send a complete ipc message every time a key is pressed. I need to think of other solutions again

dpvpro:
My solution may not be very effective, but I hope the problem will be fixed one way or another.

DreamMaoMao:
fixed

dpvpro:
Thank you

PR #891, opened 2026-05-05
feat: add blur_lockscreen option to apply blur to lock surfaces

StaszeKrk:
Adds a new `blur_lockscreen` config option (default: 1) that, when
enabled alongside the existing `blur` option, applies scenefx blur
to wlr-session-lock surfaces on each commit.

Implementation adds:
- `commit_lock_surface` listener and `lock_scene_tree` pointer to
  the Monitor struct
- `commitlocksurface` handler that calls iter_layer_scene_buffers
  on the lock scene tree
- Cleanup of both fields in destroylocksurface

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -372,6 +372,7 @@ typedef struct {
 	int32_t allow_tearing;
 	int32_t allow_shortcuts_inhibit;
 	int32_t allow_lock_transparent;
+	int32_t blur_lockscreen;
 
 	struct xkb_rule_names xkb_rules;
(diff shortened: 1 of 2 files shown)


DreamMaoMao:
I'm not inclined to do this because mainstream lock screen software itself supports blurring

StaszeKrk:
From my experience the blur doesn't work for transparent lock screen, only if the bg is a wallpaper.

StaszeKrk:
But maybe I'm just stupid and can't set it up

PR #893, opened 2026-05-06
feat: add sleep_monitor command with wake-on-input

### LINKED ISSUE #780: [Enhancement] Turn display on with mouse move/keypress
It would be good to have something like in niri.
mmsg disable monitor output - command should do something like "niri msg turn-off-monitors" does: it turns off the screen, but it turns it back on when you move mouse or press a key.

Maybe it's not a good idea to implement it into "disable monitors" command, but it would be good to see this as some additional command, like "disable_monitors_temporarily"

StaszeKrk:
Closes #780.

  Adds a sleep_monitor command that turns off a specific monitor and
  automatically re-enables it on the next keypress or mouse movement.

  Unlike disable_monitor, no explicit enable_monitor call is needed — the
  monitor wakes itself on any input event.

  Usage:
  sleep_monitor,eDP-1

  Implementation:
  - Adds wake_on_input flag to struct Monitor
  - sleep_monitor() disables the output and sets the flag
  - wakesleepingmons() is called in keypress, buttonpress, and motionnotify
  to re-enable any flagged monitors on the next real input event


DreamMaoMao:
I think the code for disabling monitor can be reused. Just need a option

DreamMaoMao:
```
swayidle \
        timeout 3 'mmsg dispatch sleep_monitor' \
        resume 'mmsg dispatch wakeup_monitor' 
```

I don't want to do any unnecessary actions in frequent event triggers. It should be uniformly handled by idle.

sleep_monitor will add in 0.15.0

PR #899, opened 2026-05-07
feat: add consume & expel for scroller stack

xtheeq:
https://github.com/user-attachments/assets/18e0a460-e485-434e-a6be-81d3acb13dc9


DreamMaoMao:
I think they should be two separate dispatch because these two parameters are not universal direction parameters

xtheeq:
> I think they should be two separate dispatch because these two parameters are not universal direction parameters

true, 
refactored to `scroller_stack_consume` & `scroller_stack_expel`

PR #904, opened 2026-05-09
feat: dwindle layout support

DreamMaoMao:
@ernestoCruz05 

I tried your dwindle layout and found it quite good. I made some optimizations. You can check it out.

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -242,6 +242,14 @@ typedef struct {
 	int32_t center_master_overspread;
 	int32_t center_when_single_stack;
 
+	/* dwindle layout */
+	int32_t dwindle_vsplit;
+	int32_t dwindle_hsplit;
+	int32_t dwindle_preserve_split;
+	int32_t dwindle_smart_split;
+	int32_t dwindle_smart_resize;
+	float dwindle_split_ratio;
+
 	uint32_t hotarea_size;
 	uint32_t hotarea_corner;
 	uint32_t enable_hotarea;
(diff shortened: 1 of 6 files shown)


ernestoCruz05:
Awesome! Im super glad you liked it. I have found an issue with the tiling preview with dwindle, where the preview predicts wrongly where a window will be tiled. I'll PR a fix in a bit.

Also, have you considered creating something like tag_carousel? So users can jump from 9->1 and 1->9 via viewleft/viewright?

ernestoCruz05:
Please see #905

DreamMaoMao:
> Please see #905 

Does it still have this problem now?

ernestoCruz05:
> > Please see #905
> 
> Does it still have this problem now?

Yeah still happening, here:

https://github.com/user-attachments/assets/a87720ec-0151-44e7-95d0-af2907e8bcbd

This is the current fix i have, should i PR it?

https://github.com/user-attachments/assets/7b641a0b-f42e-4968-81c2-4f2217dba775


I have completely removed the middle box, as i cant see a way to trigger the swap behaviour correctly? The moment we make a window float (via drag) it's deattached from the tree, so no position for the other window to swap to. Might be a skill issue.

See #906

DreamMaoMao:
https://github.com/user-attachments/assets/3743ecca-2326-4638-95ec-c36aa5486dfd

ernestoCruz05:
> 
> 
> https://github.com/user-attachments/assets/3743ecca-2326-4638-95ec-c36aa5486dfd
> 
> 
> 
> 

Oh yeah this is super cool, I also had this in mind but was hesitant as it goes a bit agaisnt the dwindle layout "mindset", but i think it's neat. How is the swapping (middle square) handled? That was the biggest mistery for me...

ernestoCruz05:
Actually let me rephrase it, i think both behaviours should be allowed and choosen by the user in the config. Most people who used dwindle are used to the 50/50 split (or 60/40) when dragging windows into other windows spaces, while your implementation is extremely cool I think most people would find the normal split (like on #906) to be more familiar? Other peoples opinion would be valuable here

DreamMaoMao:
> Actually let me rephrase it, i think both behaviours should be allowed and choosen by the user in the config. Most people who used dwindle are used to the 50/50 split (or 60/40) when dragging windows into other windows spaces, while your implementation is extremely cool I think most people would find the normal split (like on #906) to be more familiar? Other peoples opinion would be valuable here

An option can be added to decide whether to support the division in four directions. However, I have to leave for now. You can continue with your remaining work and I will merge it

Givemegitpls:
some strange dropcolor things.
If more than one window on colomn, mango use color from config. Else use default color green color
https://github.com/user-attachments/assets/7ce89d7f-2ef9-420f-b726-891190539df0

ernestoCruz05:
> some strange dropcolor things.
> 
> If more than one window on colomn, mango use color from config. Else use default color green color
> 
> https://github.com/user-attachments/assets/7ce89d7f-2ef9-420f-b726-891190539df0
> 
> 
> 
> 

This seems to be scroller? Not related to dwindle, it would be better if you created an issue

Givemegitpls:
The strange behavior of dropcolor applies to dwindle as well.
Otherwise, dwindle works perfectly.

DreamMaoMao:
> The strange behavior of dropcolor applies to dwindle as well.

you need to reopen the window after reload config.

DreamMaoMao:
?, why did I click close? How did I open it again

PR #906, opened 2026-05-09
re-fix: better tile predictions for dwindle

ernestoCruz05:
the last PR broke during the changes that were being made to dwindle


DreamMaoMao:
maybe better to directly allow dragging to support determining the cutting direction at that time

ernestoCruz05:
I'll close this one for a cleaner history?

PR #911, opened 2026-05-10
feat: add bindm (modifier-only exclusive keybind)

porl:
Added a  `bindm` config option. This is a modifier-only bind that fires on release but *only if no other bind or mousebind used the modifier during that press*.

This makes it possible to use a modifier key by itself for things like `Super` as a launcher trigger but without it firing after other combos:

    bind=SUPER,w,spawn,firefox
    bindm=SUPER,,spawn,fuzzel        # won't fire after SUPER+w
    bindm=SUPER+SHIFT,,spawn,fuzzel # works with mod combos
    bindml=SUPER,,spawn,fuzzel       # works on lock screen*

*note: I've only been able to test this in a nested environment at the moment, so whilst I did check the `bindml` line fires, I didn't check it with a real lock screen. Since it reuses the normal lock-screen logic I assume it should be okay.

Implementation: a `mod_used_for_bind` flag is set whenever a regular bind or mousebind fires while modifiers are held, and cleared on each modifier key press. `bindm` binds only fire on modifier release if the flag is unset.

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -45,6 +45,7 @@ typedef struct {
 	bool islockapply;
 	bool isreleaseapply;
 	bool ispassapply;
+	bool ismodonly;
 } KeyBinding;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


src/config/parse_config.h
```diff
@@ -2436,6 +2437,92 @@ bool parse_option(Config *config, char *key, char *value) {
 			config->key_bindings_count++;
 		}
 
+	} else if (regex_match("^bindm[lp]*$", key)) {
```
Wateir, line 2440: https://github.com/mangowm/mango/blob/cabafb2393ff71615af4959a51885c8d1eefc9a0/src/config/parse_config.h#L2346
We already have that on the config a little above, i belive it's a good idea to refractor both of this, somewhere to avoid add new modifier, like a y and then add both `bindmy` and `bindy` become hard to do and so maintain

demperor-music:
While i agree with the usecase, i don't think this is the best way to approach this as it only really solves one specific issue while implementing a whole other flag. I think a better approach would be a more generic "e" flag (exclusive), which only triggers if the bind is the only one fired while the modifier has been held down. Due to the nature of this, it would always imply `r` flag. This would allow for more configurability while also allowing to do the same thing (`binde=SUPER,SUPER_L` for SUPER-only). This way the second argument to the bind wouldn't always be empty when using this flag, which looks more in line with how other binds are configured

Just a thought for now, i might take a look and try providing some code later.

DreamMaoMao:
bind=none,Super_L,spawn,rofi -show drun

i think already support now.

src/config/parse_config.h
porl: I'm definitely happy for someone to take the idea and implement it much cleaner. I just wanted to see if it could work and hopefully get implemented in some way.

porl:
> bind=none,Super_L,spawn,rofi -show drun
> 
> i think already support now.

That seems to have the "double-spawn" problem I'm trying to fix though? Does it not behave like that for you?

porl:
> While i agree with the usecase, i don't think this is the best way to approach this as it only really solves one specific issue while implementing a whole other flag. I think a better approach would be a more generic "e" flag (exclusive), which only triggers if the bind is the only one fired while the modifier has been held down. Due to the nature of this, it would always imply `r` flag. This would allow for more configurability while also allowing to do the same thing (`binde=SUPER,SUPER_L` for SUPER-only). This way the second argument to the bind wouldn't always be empty when using this flag, which looks more in line with how other binds are configured
> 
> Just a thought for now, i might take a look and try providing some code later.

More than happy for someone to allow the same "feature" with a cleaner implementation. I assume mine is super naive but just wanted to at least get the idea working.

porl:
can confirm #918 works for me. I assume their method is cleaner so I can close this one.

Issue #886, opened 2026-05-03
Renamed packages (NixOS)


Vrayth:
Two Packages have been renamed

evaluation warning: The xorg package set has been deprecated, 'xorg.libxcb' has been renamed to 'libxcb'
evaluation warning: The xorg package set has been deprecated, 'xorg.xcbutilwm' has been renamed to 'libxcb-wm'

Thanks for your work!

GwendalBeaumont:
Probably related to #821.

PR #914, opened 2026-05-10
Updated flake.lock nixpkgs input

max-amb:
Made because project wasn't building (with `nix build`) due to the updated name for `libxcb-wm`.

Project was not building with
```nix
error: evaluation aborted with the following error message: 'lib.customisation.callPackageWith: Function called without required argument "libxcb-wm" at /nix/store/779im7kk8sq23irpb28gbbirfs2kjg3n-source/nix/default.nix:14'
```

Fixed by updating the nixpkgs input, which renamed the package to `libxcb-wm`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1750386251,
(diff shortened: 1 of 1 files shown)


max-amb:
Related to #886.

PR #918, opened 2026-05-11
Add exclusive flag feature

demperor-music:
Adds an exclusive bind flag (`e`) that marks a bind as exclusive: it only fires on modifier release if no other bind or mousebind used the modifier during that press (release flag is required). Aims to be usable for the same usecases as #911 and #693 but with a more generic approach.

Example: `bindre=SUPER,SUPER_L, [...]`
This however can also be used for binds that AREN'T modifier-only, but you just want to be exclusive for any given reason.

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -45,6 +45,7 @@ typedef struct {
 	bool islockapply;
 	bool isreleaseapply;
 	bool ispassapply;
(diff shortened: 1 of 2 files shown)


porl:
can confirm this works just as well as my version for my use case. assuming this is a cleaner solution I can probably close my pr.

PR #926, opened 2026-05-13
docs: add screenshot example binds and workflow

xtheeq:


docs/screenshot.md
```diff
@@ -0,0 +1,213 @@
...
+mangowm does not include a built-in screenshot tool. This keeps the compositor lean.
+Instead, compose your own workflow from small Wayland utilities and bind them to keys;
+
+| Tool | Purpose |
+| :--- | :--- |
+| [`grim`](https://github.com/emersion/grim) | Capture the screen or a region to a file |
+| [`slurp`](https://github.com/emersion/slurp) | Interactively select a region for `grim` |
+| [`wl-copy`](https://github.com/bugaevc/wl-clipboard) | Copy screenshots directly to the clipboard |
+| [`satty`](https://github.com/gabm/Satty) | Annotate screenshots before saving |
+| [`wayfreeze`](https://github.com/nicbk/wayfreeze) | Freeze the screen before capture |
```
Samq64, line 17: In my experience [still](https://github.com/faergeek/still) behaves better with rotated monitors. It is also more lightweight and has tagged releases.

xtheeq: i mean wayfreeze is more widely known though you are right on the tagged releases.
but these are meant to be examples, these are not the exact solutions one should follow.
I can give the alternate method, but it may be giving options fatigue for new comers.
cuz this are meant to be example of how to make screenshot with mango rather than giving a comparison of tools.

PR #932, opened 2026-05-15
feat: carousel-like behaviour when swapping tags

ernestoCruz05:
Nothing much to describe, config option that allows you to go from tag 9 -> 1 and tag 1 -> 9 with viewleft/viewright

### DIFF
--- a/src/animation/tag.h
+++ b/src/animation/tag.h
@@ -15,7 +15,11 @@ void set_tagin_animation(Monitor *m, Client *c) {
 		return;
 	}
 
-	if (m->pertag->curtag > m->pertag->prevtag) {
+	bool going_forward = m->carousel_anim_dir
+							 ? m->carousel_anim_dir > 0
+							 : m->pertag->curtag > m->pertag->prevtag;
+
+	if (going_forward) {
 
 		c->animainit_geom.x = config.tag_animation_direction == VERTICAL
 								  ? c->animation.current.x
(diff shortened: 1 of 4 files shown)


ernestoCruz05:
Oh yeah, this should be attached

https://github.com/user-attachments/assets/bba3685d-8cee-4f00-bb3c-8489e82e664f

azrtareef:
i'd   genuinely appreciate it if this feature would get merged to the wm.

i use next/prev tag binds a lot and having it stop at the last tag feels kinda awkward. With this i can just keep pressing the same bind without thinking about boundaries

also i dont really treat tags as numbered desktops, more like categories (browser, coding, chat, media etc) so looping back to the first one actually feels more natural to me

this would also help alot with sparse/dynamic workspaces because sometimes i only use a few tags actively and hitting dead ends during navigation feels weird

also, i use `viewtoright_have_client` and `viewtoleft_have_client` a lot (as i dont use dynamic tags). Adding this feature mitigates me from using 2 separate bindings as i can just cycle through the entire carousel  seamlessly without having an extra binding to go cycle through the opposite direction

sboysel:
+1 - Cycling through occupied tags, similar to cycling through windows in a stack, is a very intuitive workflow IMHO.

at5submarine:
as the people have stated above, this feature would be quite nifty and intuitive. It would substantially benefit my workflow if this feature would've been merged to mango as i frequently cycle through my tags, similarly to a stack
as well as using viewtoright with client for immaculately moving through tags with windows present on it. It would definitely enhance ones workflow in some way and it's a very good feature to be added, from my perspective. I would really be thankful to the creator if they could add this feature.

ernestoCruz05:
@DreamMaoMao hopefully these comments/reactions can you give more of an idea of the use cases of this feature

DreamMaoMao:
> @DreamMaoMao hopefully these comments/reactions can you give more of an idea of the use cases of this feature

I have responded to some reviews above and need to make some corrections before I can merge them

PR #946, opened 2026-05-19
feat: re-impl overview mode

DreamMaoMao:
dont resize window when enter overview.

https://github.com/user-attachments/assets/f0dc31ea-639a-4188-8cb6-a133316a4ca8


DreamMaoMao:
This method of forcibly scaling the buffer is unstable, so this pr is temporarily abandoned

DreamMaoMao:
Another usage method has already been adopted. You can enable it with this option

```conf
ov_no_resize=1
```

It will use the last frame before the window is triggered as a preview

PR #948, opened 2026-05-19
build(deps): add wrap files for 'scenefx' and 'wlroots'

euletheia:
## What

This allows meson to automatically fallback on fetching and building these projects in case they are not available/out-of-date on the system.
Fallback can still be avoided by passing `--wrap-mode=nofallback` to `meson setup`.

## Why

I am building Mango on Debian Testing and (in my opinion) the wrap files make the process cleaner and quicker.

The only drawback I can think of is that versions and names of these dependencies now need to be declared in two files and kept in sync (in `meson.build` and in the corresponding `*.wrap` file).

This is only a simple suggestion (I am not an expert), and I hope I am not creating too much noise.[^1]

[^1]: edited since I ended up including the `fallback` parameter in `meson.build` for better clarity.


euletheia:
For testing purpose I also specified `install_rpath` for the `mango` executable, so that I could build and run mango from a sub-folder of the repo. Maybe that can be of some use ?

```meson
diff --git a/meson.build b/meson.build
index 4f3fe6d..c9df3da 100644
--- a/meson.build
+++ b/meson.build
@@ -41,4 +41,10 @@ pcre2_dep = dependency('libpcre2-8')
 libscenefx_dep = dependency('scenefx-0.4',version: '>=0.4.1', fallback: 'libscenefx')
 
+# make subprojects libs accessible when installing with non-standard prefix
+needs_rpath = false
+foreach dep : [wlroots_dep, libscenefx_dep]
+  needs_rpath = needs_rpath or dep.type_name() == 'internal'
+endforeach
+install_rpath = needs_rpath ? join_paths(get_option('prefix'), get_option('libdir')) : ''
 
 # 获取版本信息
@@ -113,4 +119,5 @@ executable('mango',
   ],
   install : true,
+  install_rpath: install_rpath,
   c_args : c_args,
   link_args : link_args,
```

PR #951, opened 2026-05-20
fix: sync foreign toplevel visibility with tags

EzequielColungaAcha:
## Summary
- Advertise foreign toplevels to outputs only when their tags are visible on the selected tag set.
- Keep minimized clients advertised on their restore tag so taskbars can still restore them.
- Use foreign-toplevel output enter/leave events so taskbars like Waybar do not need compositor-specific IPC.

## Test plan
- Built locally through the Arch package workflow.
- Verified the PR diff is limited to `src/mango.c` and `src/ext-protocol/foreign-toplevel.h`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/ext-protocol/foreign-toplevel.h
+++ b/src/ext-protocol/foreign-toplevel.h
@@ -2,6 +2,45 @@
 
 static struct wlr_foreign_toplevel_manager_v1 *foreign_toplevel_manager;
 
+static bool foreign_toplevel_should_enter_output(Client *c) {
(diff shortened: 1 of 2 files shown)


EzequielColungaAcha:
Basically, used with wlr/taskbar it shows the windows open on the current tag only, making the click to minimize/maximize work as in DWM.

DreamMaoMao:
We cannot do this because this protocol is not merely an example to support this function; it is a general protocol and will have other application scenarios

PR #967, opened 2026-05-24
feat: better ipc support

DreamMaoMao:
useag
socket path:
```
echo $MANGO_INSTANCE_SIGNATURE
```

```bash
mmsg get keymode
mmsg get keyboardlayout
mmsg get all-clients
mmsg get all-tags
mmsg get all-monitor

mmsg get client 5
mmsg get tag eDP-1 9
mmsg get monitor eDP-1

mmsg dispatch movewin,1000,100

mmsg dispatch exchange_client,left client,4
```

watch mode you need to change the get to watch, for example
```bash
mmsg watch all-tags
```


DreamMaoMao:
Note that once this pr format matures, the outdated dwl ipc protocol will be completely removed. We hope that the relevant ecosystem tools can help update to the new ipc.

prankstr:
One question about overview mode.. Mangos `ext-workspace` implementation shows overview as a special workspace with tag `0` and the name `overview` which replaces the regular 1-9 tags. The new IPC looks closer to dwl-ipc here where overview appears as tags `1-9` all being active.

Would it be possible to expose overview explicitly in the new IPC or mirror the `ext-workspace` representation there too?

That would let panels/bars show a single `overview` item while using only the new IPC instead of having to combine IPC and `ext-workspace` state.

DreamMaoMao:
> Would it be possible to expose overview explicitly in the new IPC or mirror the `ext-workspace` representation there too?

i will set it to 0 too .

prankstr:
> i will set it to 0 too .

Great! Another thing that would be nice is to be able focus a client by its `id`. Since `all-clients` exposes stable client ids, a command like `focus client <id>` would let panels implement click-to-focus.

DreamMaoMao:
> > i will set it to 0 too .我也会把它设为0。
> 
> Great! Another thing that would be nice is to be able focus a client by its `id`. Since `all-clients` exposes stable client ids, a command like `focus client <id>` 


```
 mmsg dispatch focusid client,3

```

Wateir:
> echo $MANGO_INSTANCE_SIGNATURE

This can be a better solution to handle my implmentation for https://github.com/mangowm/mango/pull/848

mmsg/mmsg.c
```diff
@@ -1,754 +1,73 @@
...
-static void dwl_ipc_output_active(void *data,
-								  struct zdwl_ipc_output_v2 *dwl_ipc_output,
-								  uint32_t active) {
-	if (!oflag) {
-		if (mode & SET && !output_name && active)
-			output_name = strdup(data);
-		return;
+	const char *socket_path = getenv("MANGO_INSTANCE_SIGNATURE");
+	if (!socket_path) {
+		fprintf(stderr, "Error: MANGO_INSTANCE_SIGNATURE not set.\n");
```
Wateir: "Error: MANGO_INSTANCE_SIGNATURE not set. Do mango is running ?\n"
Seems a better sentence for end user who can't not be award of what is $MANGO_INSTANCE_SIGNATURE

sewnie:
Why not expose mango IPC over a Wayland protocol, allowing normal applications to access it? The original protocol originally was based on top of the dwl unstable protocol with changes. I believe both are still possible? If so, why? Two ways to access the IPC where only one is specific to mango is quite strange.

DreamMaoMao:
> Why not expose mango IPC over a Wayland protocol, allowing normal applications to access it? The original protocol originally was based on top of the dwl unstable protocol with changes. I believe both are still possible? If so, why? Two ways to access the IPC where only one is specific to mango is quite strange.


I will remove support for the dwl protocol later,

Even the slightest modification in this approach will completely break the entire protocol, forcing all clients to update the protocol once. Moreover, the wayland extension protocol has a limit on the length of the string.

So even for sway, they can only choose an independent ipc.

sewnie:
> the wayland extension protocol has a limit on the length of the string.

Whats the usecase for this in Mango? Surely youre not communicating stringed json to a Wayland client.

DreamMaoMao:
> Whats the usecase for this in Mango? Surely youre not communicating stringed json to a Wayland client.在 Mango 

we need get all client message by client, You can see there are a large number of requests on the issue.

MoeMonsuta:
Sounds great, I hope it can support opening specific apps on auto start in a specific tag! The hard code window rule workaround is not ideal for me.

Issue #981, opened 2026-05-26
Add a keybinding for promoting a window to Master window in the Dwindle layout


TPhakisi:
The first window to spawn in the dwindle layout is usually the master window and it takes half the screen when other windows spawn. I'd like to have a keybinding to promote any window in the dwindle tree to be the master instead of manually moving it around....as one can in hyprland. It helps with ease of use for the layout.

Thank

ernestoCruz05:
Dwindle does not follow the "Master-Stack" behaviour, as all windows are leafs in a BSP, so there is no Master window.

I am not familiar with this functionality in Hyprland and looking in the Dwindle Layout hyprland wiki page i don't see anything related to that. Can you clarify more? Maybe give the keybind you are talking about.


Just from the top of my head, this could be implemmented but that would require completely mutating the tree to move one leaf to the top and every other leaf into the second branch... it would be very unstable

TPhakisi:
@ernestoCruz05 it is the `movetoroot` layout message under the dwindle layout on their wiki/documentation

ernestoCruz05:
> @ernestoCruz05 it is the `movetoroot` layout message under the dwindle layout on their wiki/documentation
> 

Thank you, i will look into it

ernestoCruz05:
Also, i can implement it but its still up to @DreamMaoMao to decide wether it should be merged or not

PR #983, opened 2026-05-26
feat: dwindle movetoroot dispatch

ernestoCruz05:
See #981


ernestoCruz05:
Forgot to attach:

https://github.com/user-attachments/assets/eb664ae5-d45a-4837-a4a1-d0b7909204f1

The first few times im using unstable, after that im using this keybind:
bind=SUPER,p,movetoroot
which does not have unstable

PR #1009, opened 2026-06-02
Feature: Add optional titlebars

xdavidel:
Don't know if this an interest of this project but I've created an implementation of titlebars for clients with dragging support and a closing button.

I don't think this should be accepted as is even if it is a feature that is planned, but I'm willing to work on it to perfect it if necessary.

### DIFF
--- a/src/animation/client.h
+++ b/src/animation/client.h
@@ -539,7 +539,7 @@ void apply_titlebar(Client *c, int32_t rect_width,
 }
 
 void update_titlebar_hover(Client *c) {
-	if (!c || !c->titlebar_close || !c->titlebar_bg)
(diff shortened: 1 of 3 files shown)


DreamMaoMao:
This is not an correct way to draw. There is a module for drawing the title bar in the next branch

https://github.com/mangowm/mango/pull/1079

xdavidel:
> This is not an correct way to draw. There is a module for drawing the title bar in the next branch
> 
> #1079

I see. Is that the same a drawing the group bar?
In any case is there a milestone / planned features / other way to track ongoing work?
Maybe I missed it...

Thanks for your hard work.

PR #1022, opened 2026-06-06
feat: move clients

ernestoCruz05:
I really like this feature, especially with dwindle which is the layout it makes the most sense in

https://github.com/user-attachments/assets/6dc3e26a-4913-43d6-93d2-394f88717ef1

Binding goes something like this:
bind=SUPER+ALT,Up,move_client,up
bind=SUPER+ALT,Down,move_client,down
bind=SUPER+ALT,Left,move_client,left
bind=SUPER+ALT,Right,move_client,right


DreamMaoMao:
done

PR #1024, opened 2026-06-06
build: rename binaries and paths to allow side-by-side installation

CtrlLuzzio:
Changed the binary names to mango-wl and mmsg-wl and the configuration paths to mango-wl/ to allow parallel installations of mango main branch and wl-roots only.

### DIFF
--- a/docs/ipc.md
+++ b/docs/ipc.md
@@ -24,6 +24,7 @@ description: Control mangowm programmatically using mmsg.
 | Command | Description |
 | :--- | :--- |
 | `get version` | Returns the current version of the compositor. |
+| `get cursorpos` | Returns the global pointer position (`x`, `y`) and the monitor under it. |
 | `get keymode` | Returns the current active keyboard mode (e.g., normal, insert). |
 | `get keyboardlayout` | Returns the active XKB layout (abbreviated). |
 | `get monitor <name>` | Returns full JSON details for a specific monitor. |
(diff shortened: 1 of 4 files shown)


Yappaholic:
You probably did not rebase before changing the source code and it can't be reviewed since you pull 234 changes from the upstream.

CtrlLuzzio:
Actually I forked before the changes I just forgot to update my fork, I'll do it  when I get back from work.

CtrlLuzzio:
@Yappaholic done

Yappaholic:
Is it really necessary to also rename config and man paths? Would the default ones mess up the package manager database?

CtrlLuzzio:
Config path is not really necessary since if you have a base mango config and load mango-wl-only  it would use the config but ignore the invalid parameters. However, I thought it was best to have separate config paths since it gives more control to the user.

About the man, to my knowledge, leaving the default names would mess up the package manager because it would try to claim ownership of the same files if an user installs both sessions at the same time (which is the point of this change).

Yappaholic:
> Config path is not really necessary since if you have a base mango config and load mango-wl-only it would use the config but ignore the invalid parameters. However, I thought it was best to have separate config paths since it gives more control to the user.
> 
> About the man, to my knowledge, leaving the default names would mess up the package manager because it would try to claim ownership of the same files if an user installs both sessions at the same time (which is the point of this change).

In this case I would advise against adding new config path to reduce confusion, everything else is fine, but I'm not using wl-only, so can't vouch for this change

PR #1028, opened 2026-06-09
nix: drop flake-parts and add support for systems that don't use flakes

pengolord:
Some users, myself included, don't use flakes, and I'd still like to be able to get the nixos module, home-manager module, and nightly package without having to write everything myself.

This pr adds a `default.nix` file to the root of the repo, meaning you can do this to install mango nightly:
```nix
{ pkgs, ... }:
let
  commit = ""; # commit of this repo to download from
  sha256 = ""; # hash of the repo at that commit, to ensure reproducibility

  src = fetchTarball {
    url = "https://github.com/mangowm/mango/archive/${commit}.tar.gz";
    inherit sha256;
  };
  
  mango = import src {
    # inherit pkgs; # optional, equivalent of `inputs.mangowm.nixpkgs.follow = "nixpkgs";`
  };
in {
  environment.systemPackages = [
    mango.package
  ];
}
```
or, if using a source pinning tool such as tack, npins, nvfetcher, or lon:
```nix
{ pkgs, sources, ... }:
let
  mango = import sources.mango {
    # inherit pkgs; # optional, equivalent of `inputs.mangowm.nixpkgs.follow = "nixpkgs";`
  };
in {
  environment.systemPackages = [
    mango.package
  ];
}
```
similar things can be done to install the nixos module, home-manager module, and overlay.

if needed, we could move the `default.nix` into the `./nix` subdirectory, which might clean up the root but it'd make installation look a bit strange:
```nix
mango = import "${sources.mango}/nix" {
  # inherit pkgs; # optional, equivalent of `inputs.mangowm.nixpkgs.follow = "nixpkgs";`
};
```

I'm more than open to any suggestions for changes or improvements, and I also think it needs a bit more testing. marking this as a draft for now.


Yappaholic:
Building package with `nix build github:pengolord/mangowm#packages.x86_64-linux.default` is working, next I'll try to replace the input and check if system/hm modules work.

Yappaholic:
Rebuilding the system with:
```nix
inputs.mangowm = {
  url = "github:pengolord/mangowm";
  inputs.nixpkgs.follows = "nixpkgs";
};
```
works without issues both for system and home-manger modules.

Would be nice if more people could test this, but I'm happy with the current result.

Issue #560, opened 2026-01-07
Vertical tag animation leaves artefacts in blurred waybar


bgmeulem:
## Info

mango version 

0.10.9(23d550e)

When setting 

```
tag_animation_direction=0
```
The vertical animations leave some artefacts in my blurred waybar. 
After the vertical animation has finished, horizontal animations trigger different kinds of artefacts.
It looks like the coordinates used in blur calculation do not get refreshed fast enough, or at least aren't refreshed when the animation finishes, leaving some parts of the waybar to include pixels in blurring that it really shouldn't.

Other settings:
```
blur=1
blur_layer=1
blur_optimized=0
blur_params_num_passes = 2
blur_params_radius = 5
blur_params_noise = 0.02
blur_params_brightness = 0.9
blur_params_contrast = 0.9
blur_params_saturation = 1.2
```

https://github.com/user-attachments/assets/de6bc1bf-f93b-4d1e-bd0f-04640185c893

Vishvesh1542:
https://github.com/user-attachments/assets/2a6d2721-aefa-4e85-a53f-a0c00574f131

Also when you move windows to a corner, the blur suddenly starts considering the borders in which makes it look really bad.

Also when animating walker, there is a sudden shift when the blur sort of changes. ( I haven't set up walker yet so I'm not sure if the issue is just that )

I believe these two issues can be fixed together

bgmeulem:
Some additional information that could be useful:
- This does not happen with optimized blur
- This happens with other status bars as well. 

I've been testing [`vibepanel`](https://github.com/prankstr/vibepanel) and this gives the same artifacts in the status bar. In addition, `vibepanel` has similar artifacts for buttons, but only when overlayed on the wallpaper. When overlayed with other clients, it behaves perfectly fine. 

Video in this issue: https://github.com/prankstr/vibepanel/issues/70

bgmeulem:
Also interesting, it seems to happen with elements that contain both opaque and transparent elements, but not transparent-only elements. I could successfully remove such artefacts from some layers by setting opacity of everything (text, icons, background) to 0.99

https://github.com/prankstr/vibepanel/issues/70#issuecomment-4063869669

bgmeulem:
Apparently, any semi-transparent panel with blur (that contains opaque elements) will do. here's just two terminals:

https://github.com/user-attachments/assets/a70bb4b4-4eec-4363-afb0-e9e92c70f5ce

[...]

PR #1044, opened 2026-06-13
fix: dirty the blur buffer during and after animation/resize

bgmeulem:
Fixes issue #560
On my machine (integrated graphics: AMD Ryzen 5 7640U w/ Radeon 760M Graphics), I have rendering artifacts when two blurred surfaces pass over each other.
`scenefx` does not automatically dirty the blur buffer for you (see https://github.com/wlrfx/scenefx/issues/94), so it is currently the job of the compositor to do so.

This commit manually marks the blur buffer as dirty during client resizing and animation, fixing the artifact issue on my end.

I added a new flag `need_blur_dirty` in `rendermon` that marks the blur buffer as dirty is an animation is playing. I could have used `need_more_frames` as well, but then the last frame would not dirty the buffer, when it should.

Open questions:
- why these artifacts only seem to happen on some machines. Not many people seem to have the same issue, from what I can tell? Maybe they simply don't use a lot of blur?
- is this the best way of implementing this?

Open to any feedback - this is my first PR here

### DIFF
--- a/src/animation/client.h
+++ b/src/animation/client.h
@@ -1243,6 +1243,10 @@ void resize(Client *c, struct wlr_box geo, int32_t interact) {
 	if (!c->mon)
 		return;
 
+	if (config.blur && c->mon->blur) {
+		wlr_scene_optimized_blur_mark_dirty(c->mon->blur);
+	}
+
 	c->need_output_flush = true;
 	c->dirty = true;
 
--- a/src/mango.c
+++ b/src/mango.c
@@ -5045,6 +5045,7 @@ void rendermon(struct wl_listener *listener, void *data) {
 	bool frame_allow_tearing = false;
 	struct timespec now;
 	bool need_more_frames = false;
+	bool need_blur_dirty = false;  // dirty the blur buffer
 
 	if (session && !session->active) {
 		return;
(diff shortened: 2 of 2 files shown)


justinesmithies:
I'm not sure if this is the same issue but when I select overview mode then come out of it the shadows on my waybar start kind of rolling. They don't move postition but you can see darker grey bars rolling up until I click on waybar. Nothing else is affected and it only happens when I open and close the overview mode to the shadows on waybar.

bgmeulem:
> I'm not sure if this is the same issue but when I select overview mode then come out of it the shadows on my waybar start kind of rolling. They don't move postition but you can see darker grey bars rolling up until I click on waybar. Nothing else is affected and it only happens when I open and close the overview mode to the shadows on waybar.

Do you have blur enabled on waybar? If so, it could. Feel free to test this PR and see if that fixes it. 

Also, maybe we should move this discussion to the issue I linked. I suspect the discussion section of a PR is reserved for the PR itself? Correct me if I'm wrong though, I'm new to this.

DreamMaoMao:
>scenefx does not automatically dirty the blur buffer for you (see https://github.com/wlrfx/scenefx/issues/94), so it is currently the job of the compositor to do so.

The meaning here is that it only needs to be triggered when the background changes.

If you are not using blur optimization, there is no need for us to calculate the damage, there should some issue in scenefx damage detect.

PR #1075, opened 2026-06-18
feat: monocle tab bar toggle

coil111:
the monocle tab bar introduced for multi-window monocle is always on and has no option to be disabled, the only related knob (tab_bar_height) is clamped to a minimum of 5, so it cannot be turned off through configuration

added a tab_bar_enable boolean (defaults to 1 to preserve current behaviour) 
when set to 0 falls back to classic monocle behaviour

also initialized fc to NULL
it was used uninitialized and if the fstack scan found no match it would  feed garbage to client_tile_resize


DreamMaoMao:
you can set bar height to 0 now.

PR #1078, opened 2026-06-19
feat(ipc): add 'get cursorpos' 

ernestoCruz05:
I think the cursorpos will be really useful for making bash scripts while the get option is a nice QoL for checking in scripts too, but i do realize that it increases the ammount of lines needed when adding new config values as it is implemented as a read counterpart to parse_option, so adding a new config value now means adding it to the read block too. 

There's a cleaner long-term solution refactoring the options into a single table so that parsing and reading are driven by one definition, killing the double-maintenance entirely. But that's a bigger, small-scope refactor of parse_option, and I didn't want to rewrite a core working function uninvited. Happy to do it if you'd want that, but I understand if you'd rather not have someone else reshaping that part.


ignition-ctrl:
There's also #592 for this as well.

Issue #735, opened 2026-03-05
Plugins Support


NikunjK8:
One of the biggest reason why Hyprland is so popular today is because of it's huge plugins ecosystem, consisting hundreds of both official & unofficial plugins. Since mango is getting a lot of traction, I wanted to ask if this is something you can consider for future. Is building a plugins API in scope of this project?

This would also mean you don't have to add many features you don't want in Mango itself. People can get those functionalities through plugins & plugin devs will be the maintainers. I'm mentioning this because you said you want to keep mango stable and only add practical features (I'm in favor of this decision since many programs becomes extremely bloated overtime).

Having a robust plugin ecosystem will give users both:
1. A stable & maintainable base with all the necessary functionality.
2. Endless possibilities & options to further shape their workflow according to their needs.

glxguard:
Yeah, this also will give an option for MangoWM to stay lightweight, which is good

NikunjK8:
> Yeah, this also will give an option for MangoWM to stay lightweight, which is good

True. I switched exactly because Hyprland was becoming too bloated and unstable to use. Breaking changes in every release drove me insane.

zPhoeniqz:
I like the idea. How would this be implemented? I think going the hyprland way would not necessarily be a smart choice here - hyprland comes with an entire package manager (`hyprpm`) for plugins, which I think is a bit overkill and contradicts the lightweight design goal of mango.

NikunjK8:
Either natively adding plugins support in mango (which will increase the codebase size a bit but it's better than keep adding features yourself rather than letting plugins do a lot of user-specific customizations). Or if separate package is used than make it optional rather than hard dependency.

kociap:
You might not like the answer, but we all what the right way here is - lua.

NikunjK8:
@kociap I'm a lua supporter myself #1321

PR #1079, opened 2026-06-21
mango 0.15.0

DreamMaoMao:
- [x] **window group support**
- [x]  **hdr support**
- [x] **overview label jump mode**
- [x]  **vrr only fullscreen**
- [x] **shield for screen record.**
- [x] **capturing/casting individual windows**

Since most people prefer the original monocle mode, the tab bar has been removed.

https://github.com/user-attachments/assets/f6f5fff4-d271-4099-92ac-af5f62102f94


xtheeq:
Thats super dope!! :fire:

xtheeq:
is it possible reusing focusdir/focusstack instead of a new bind groupfocus??

DreamMaoMao:
> is it possible reusing focusdir/focusstack instead of a new bind groupfocus??

How to distinguish them?

xtheeq:
Not distinguish but use focusstack the same way monocole did.
while in a group it goes through the group, at the start/end it falls through to the stack.

xtheeq:
> Not distinguish but use focusstack the same way monocole did. while in a group it goes through the group, at the start/end it falls through to the stack.

Example flow:

https://github.com/user-attachments/assets/768b5d3e-e50e-4da6-b3aa-86ad4d111367

config example:
```
bind=SUPER,n,focusstack,next
bind=SUPER,p,focusstack,prev
```

DreamMaoMao:
> > Not distinguish but use focusstack the same way monocole did. while in a group it goes through the group, at the start/end it falls through to the stack.
> 
> Example flow:
> 
>  20260621194641.mp4 
> config example:
> 
> ```
> bind=SUPER,n,focusstack,next
> bind=SUPER,p,focusstack,prev
> ```

I'm not quite sure how the person who made the request before used it. This might need to be discussed

xtheeq:
https://github.com/user-attachments/assets/b6cf314d-c6f1-4593-af66-603e7e258e80

I've been also trying to make a single focus bind unified to work for all.
This if done properly can be much simpler for usage. consider this if its possilbe.

xtheeq:
Noticed that in overview mode doing groupjoin / groupleave makes crash and weird behaviors.

ernestoCruz05:
This looks really cool! One question tho, to move a window from the group to another group, or to another monitor, will I need to expel it from the group then move?

DreamMaoMao:
> Noticed that in overview mode doing groupjoin / groupleave makes crash and weird behaviors.

This operation has been prohibited.

DreamMaoMao:
> This looks really cool! One question tho, to move a window from the group to another group, or to another monitor, will I need to expel it from the group then move?

it work now.

DreamMaoMao:
> > Not distinguish but use focusstack the same way monocole did. while in a group it goes through the group, at the start/end it falls through to the stack
> Example flow:  流程示例：
> 
>  20260621194641.mp4 
> config example:  配置示例：
> 
> ```
> bind=SUPER,n,focusstack,next
> bind=SUPER,p,focusstack,prev
> ```


https://github.com/user-attachments/assets/12d11674-08ee-4ca3-ac14-9cbbb416aa3d

This seems to become very counterintuitive. Without careful prediction, it's hard to know what the next window will switch to

SDG-Den:
would it be possible to add the option to "auto-group" the stack in deck mode and monocle mode? (replicating the previous monocle tab bar behaviour)

I personally really liked the monocle tab bar (to the point where i've made my own implementations for it *multiple* times for it across multiple window managers), and actually use it daily now! it'd be nice to keep it as an option by integrating it into the group system. 

 a way to quickly add all windows on the current view to a group would also be nice (with optional synctag for multi-monitor)

xtheeq:
> This seems to become very counterintuitive. Without careful prediction, it's hard to know what the next window will switch to

I'm not sure why you feel this is counterintuitive, for me while testing i feel this currently preserves the previous behaviours as well as doesn't require an extra bind which is the most important one.

xtheeq:
A idea / request:
A separate indicator bar an alternative to the tabs.
this would be minimal in the screen and doesn't take up quite large space.
example from niri (see left side in center):
image

wzykubek:
I think that this feature although useful for *some* people introduces a lot of confusion for current users. That makes monocle and deck layouts basically useless. I know that people are not forced to use this feature, however they are not forced to use mango at all. All current layouts - except for dwindle - follow the philosophy of being dynamic, not manual unlike this tabbing method. 

That makes sense, because mango is a dwl (dynamic window *manager*) fork. Manual tabbing and splitting is known from i3/sway. I think that we shouldn't try to replicate every possible layout/behaviour from other popular compositors just because some users like that. In my opinion, we should leave manual tiling to sway and focus on dynamic tiling only.

That being said, I think that previous visual representation for monocle layout should be available (disabled by default) and also introduced for deck layout.

Mango in the current state is a fair alternative for previous dwm users, who liked the idea of dynamic tiling and minimal set of features, but not necessarily the idea of manual patching. 

Edit: Maybe this kind of feature is a good candidate for being a plugin as proposed in #735.

DreamMaoMao:
> A idea / request: A separate indicator bar an alternative to the tabs. this would be minimal in the screen and doesn't take up quite large space. example from niri (see left side in center): image

DreamMaoMao:
> would it be possible to add the option to "auto-group" the stack in deck mode and monocle mode? (replicating the previous monocle tab bar behaviour)
> 
> I personally really liked the monocle tab bar (to the point where i've made my own implementations for it _multiple_ times for it across multiple window managers), and actually use it daily now! it'd be nice to keep it as an option by integrating it into the group system.
> 
> a way to quickly add all windows on the current view to a group would also be nice (with optional synctag for multi-monitor)

Sorry, you can raise another issue. I'll see how many people need it because it involves too many special logics in the code architecture and the cost of embedding this function in the code is relatively high.

DreamMaoMao:
> I think that this feature although useful for _some_ people introduces a lot of confusion for current users. That makes monocle and deck layouts basically useless. I know that people are not forced to use this feature, however they are not forced to use mango at all. All current layouts - except for dwindle - follow the philosophy of being dynamic, not manual unlike this tabbing method.
> 
> That makes sense, because mango is a dwl (dynamic window _manager_) fork. Manual tabbing and splitting is known from i3/sway. I think that we shouldn't try to replicate every possible layout/behaviour from other popular compositors just because some users like that. In my opinion, we should leave manual tiling to sway and focus on dynamic tiling only.
> 

I do not follow the suckless principle. As long as this feature is not highly coupled with other features, is effective in certain working scenarios, and many people request it, I will consider adding it.

Besides, I'm not inclined to implement plugins because, based on the experience of other compositor, there are very few users who can write plugins, and few of these few plugins have been well maintained since then

DreamMaoMao:
> > A idea / request: A separate indicator bar an alternative to the tabs. this would be minimal in the screen and doesn't take up quite large space. example from niri (see left side in center): image 
> why not set the group bar small?为什么不把小组规模缩小一些呢？
> 
> image

DreamMaoMao:
> > This seems to become very counterintuitive. Without careful prediction, it's hard to know what the next window will switch to
> 
> I'm not sure why you feel this is counterintuitive, for me while testing i feel this currently preserves the previous behaviours as well as doesn't require an extra bind which is the most important one.

for example:
I operated in window 1, and in 2 and 3, I placed a reference window that I wouldn't view at the same time.

I think the purpose of a group is that I only want to display a certain one within the group at present, so I will put it in the group. However, the shortcut key traversal will change the current focus within the group, which violates this awareness. If the window is to be fully traversed from the very beginning, then why put it in the group instead of outside.



https://github.com/user-attachments/assets/d2c00aa6-9df8-4370-90dc-8806bcef6b72

For instance, if I only want to view the contents of Windows 1 and 2 at the same time, or the contents of Windows 1 and 3, and if I traverse them all, then when I return to 1 to work, I can only view Windows 1 and 3. Windows 1 and 2 will never appear simultaneously unless I traverse them all in the opposite direction again.


Could you describe the scenarios in which you use the group? Why is it necessary to put the window into the group?

xtheeq:
> > A idea / request: A separate indicator bar an alternative to the tabs. this would be minimal in the screen and doesn't take up quite large space. example from niri (see left side in center): image 
> why not set the group small?
> 
> image

I'm not sure how did you set this. I only managed to shrink the size make the colors adjust but I don't feel its nice.
could you share the config on how you set this?

DreamMaoMao:
> > > A idea / request: A separate indicator bar an alternative to the tabs. this would be minimal in the screen and doesn't take up quite large space. example from niri (see left side in center): image > 
> > 
> > why not set the group small?为什么不把小组规模缩小一些呢？
> > image
> 
> I'm not sure how did you set this. I only managed to shrink the size make the colors adjust but I don't feel its nice.

```
group_bar_height = 10
group_bar_decorate_fg_color=0x505050ff
group_bar_decorate_bg_color=0x505050ff
group_bar_decorate_focus_fg_color=0x7FC8FFff
group_bar_decorate_focus_bg_color=0x7FC8FFff
group_bar_decorate_border_color=0x00000000
group_bar_decorate_border_width = 2
group_bar_decorate_corner_radius = 0
group_bar_decorate_padding_x = 0
group_bar_decorate_padding_y = 0
```

xtheeq:
> I think the purpose of a group is that I only want to display a certain one within the group at present, so I will put it in the group. However, the shortcut key traversal will change the current focus within the group, which violates this awareness. If the window is to be fully traversed from the very beginning, then why put it in the group instead of outside.
> 
>  2026-06-26_06-53-43.mp4 
> For instance, if I only want to view the contents of Windows 1 and 2 at the same time, or the contents of Windows 1 and 3, and if I traverse them all, then when I return to 1 to work, I can only view Windows 1 and 3. Windows 1 and 2 will never appear simultaneously unless I traverse them all in the opposite direction again.
> 
> Could you describe the scenarios in which you use the group? Why is it necessary to put the window into the group?

I don't use focusstack as my primary navigation because its slow moving through all the windows whereas with focusdir I can navigate faster.
I would use focusstack only when being inside the group since focusdir doesn't have any effect.
The violating case you gave is true when users only use focusstack for their navigation and not focusdir.
When i want to focus different window outside group I don't use focusstack rather focusdir so that focus is not changed.
I'm not a dwm/dwl & never have been. So this might be biased but I'm not sure.

The main concern for me is keybind crisis.
Most of the modifiers for me is already occupied.
using multiple modifiers combinations isn't really convenient.

Having a bind that's case specific and useless outside that case, while another bind does the same thing, is why I suggested unifying them.

xtheeq:
There's some issues on reload to take effect on the group bar.
While in a group the effect only takes on the focused group bar and not for all.

xtheeq:
Noticed that while the bar height is too small the text overflows the bar.
is there a way to handle this?
image

DreamMaoMao:
> Noticed that while the bar height is too small the text overflows the bar. is there a way to handle this? image

group_bar_decorate_fg_color=0x00000000
group_bar_decorate_focus_fg_color=0x00000000

Set transparency

xtheeq:
> There's some issues on reload to take effect on the group bar. While in a group the effect only takes on the focused group bar and not for all.

Testing seems to be checked.
This issue is still present though.

https://github.com/user-attachments/assets/2bf8ac3c-4d97-4e8c-ae61-23ad1aa91ca7

DreamMaoMao:
> > There's some issues on reload to take effect on the group bar. While in a group the effect only takes on the focused group bar and not for all.
> 
> Testing seems to be checked. This issue is still present though.
> 
>  20260626210112.mp4

I miss this. should work now

georgemoodier:
Does this PR implement capturing/casting individual windows? (Wlroots20.0 has support for this)

DreamMaoMao:
> Does this PR implement capturing/casting individual windows? (Wlroots20.0 has support for this)

yes, it does

DreamMaoMao:
> > I think the purpose of a group is that I only want to display a certain one within the group at present, so I will put it in the group. However, the shortcut key traversal will change the current focus within the group, which violates this awareness. If the window is to be fully traversed from the very beginning, then why put it in the group instead of outside.
> > 2026-06-26_06-53-43.mp4
> > For instance, if I only want to view the contents of Windows 1 and 2 at the same time, or the contents of Windows 1 and 3, and if I traverse them all, then when I return to 1 to work, I can only view Windows 1 and 3. Windows 1 and 2 will never appear simultaneously unless I traverse them all in the opposite direction again.
> > Could you describe the scenarios in which you use the group? Why is it necessary to put the window into the group?
> 

Based on my previous communication with others, most people do not use both focusdir and focusstack at the same time; they only use one of them.

BlackCherryCat:
Hi, I have a question about this amazing release, it almost adds everything I want to have in mango, but the shield also work in layers like notifications?

DreamMaoMao:
> Hi, I have a question about this amazing release, it almost adds everything I want to have in mango, but the shield also work in layers like notifications?

I don't understand what this means.

georgemoodier:
> Hi, I have a question about this amazing release, it almost adds everything I want to have in mango, but the shield also work in layers like notifications?

looking at the docs on the branch, only window clients have `shield_when_capture` listed as a parameter

DreamMaoMao:
> > Hi, I have a question about this amazing release, it almost adds everything I want to have in mango, but the shield also work in layers like notifications?
> 
> looking at the docs on the branch, only window clients have `shield_when_capture` listed as a parameter

To be honest, doing this doesn't make much sense because the actual size of most layer applications is not the size they display. On the surface, it just looks like a very small notification bar, but in fact, there is a huge hidden transparent panel behind it. Doing this will only cover the entire screen.

BlackCherryCat:
For me at least works perfectly for blocking dunst and mako notifications in screensharing in other compositors, that's why I wanted it🥺

DreamMaoMao:
> For me at least works perfectly for blocking dunst and mako notifications in screensharing in other compositors, that's why I wanted it🥺

I don't think so. Could you provide the following videos of the swaync pop-up notifications shield on other compositor?

BlackCherryCat:
image
In the case of swaync its more or less bad but for dunst:
image
its perfect, thats why I use the shield in notifications

DreamMaoMao:
@georgemoodier done.

xtheeq:
> Based on my previous communication with others, most people do not use both focusdir and focusstack at the same time; they only use one of them.

That's actually the reason I ask for the unification of binds. Since most won't use many binds adding a new bind for a similar action means it would be quite hard to manage binds.

I don't ask for removing group focus but just consider allowing focusstack to go through groups, like the implementation you did but removed currently.
If you think focusstack iterating through groups breaks pattern maybe allow with a config.

xtheeq:
A small feature request to expose toplevel window identifier in ipc.
This would be helpful for taking screenshots of hidden windows using grim like `grim -T <identifier>`.

xtheeq:
Thank YOU :heart:

njdom24:
This is really cool! I'd like to test on Nix, but I think the flake needs an update.

Soliprem:
@njdom24 would you mind testing #1126 ? Runs fine for me with it

njdom24:
Not sure if I'm doing anything wrong, but
```
env=WLR_RENDERER,vulkan
monitorrule=name:DP-1,hdr:1
```

won't make my monitor enter HDR mode. I'm also not seeing the color management protocol exposed, which would prevent apps from being aware they should run in HDR mode.

(Please disregard if this is a WIP)

Soliprem:
same here. Maybe it's a dependency problem with my nix pr? Don't see how that could be since I just bumped the dependencies, though. Sounds very unlikely

DreamMaoMao:
> Not sure if I'm doing anything wrong, but
> 
> ```
> env=WLR_RENDERER,vulkan
> monitorrule=name:DP-1,hdr:1
> ```
> 
> won't make my monitor enter HDR mode. I'm also not seeing the color management protocol exposed, which would prevent apps from being aware they should run in HDR mode.
> 
> (Please disregard if this is a WIP)

could you try this branch?

https://github.com/mangowm/mango/pull/1100

njdom24:
> could you try this branch?
> 
> #1100

At least on Nix with the flake updated like above, I see the same issue: No color management protocol available, and I'm unable to switch my displays to HDR.

DreamMaoMao:
> > could you try this branch?
> > #1100
> 
> At least on Nix with the flake updated like above, I see the same issue: No color management protocol available, and I'm unable to switch my displays to HDR.

You probably failed to switch to that branch. You can confirm by outputting mango -v. Currently, hdr can only be enabled in this branch. Although the main branch has the logic to support hdr, since the main branch relies on scenefx, the vulkan backend cannot be enabled temporarily

njdom24:
I tried switching my flake's url to 
```
    mangowm= {
      #url = "github:mangowm/mango";
      url = "git+https://github.com/mangowm/mango?ref=wl-only&submodules=1";
      inputs.nixpkgs.follows = "nixpkgs";
    };
```

and running `WLR_RENDER=vulkan mango`, and cloning the repo, switching to wl-only, and running `WLR_RENDERER=vulkan nix run .\#packages.x86_64-linux.mango`. Same behavior. mango -v outputs mango 0.14.4(release) in both cases, presumably due to `meson.build` reporting that version, and `nix` building in a containerized environment.

PR #1088, opened 2026-06-22
fix: use output scale for tab bar text rendering on HiDPI displays

emilia-miki:
mango_tab_bar_node_update was called with a hardcoded scale of 1.0 in both client_add_tab_bar_node and updatetitle.  On HiDPI outputs the compositor scales the resulting 1x buffer up, producing blurry text.

Pass the client's output scale (wlr_output->scale) instead, falling back to 1.0 when the client has no monitor assigned yet.


DreamMaoMao:
https://github.com/mangowm/mango/pull/1079

It has already been replaced here

Issue #937, opened 2026-05-15
todo list


DreamMaoMao:
- [x] **Implement a more robust and independent IPC mechanism**
- [x] **Achieve smoother gesture controls**
- [x] **Refactor the project into a more modular structure**
- [x] **tablet support**
- [x] **touchscreen support**
- [x] **floating tag support**
- [x] **HDR support**
- [x] **Introduce window decorations to achieve window grouping**

praefrbx:
Just add the ignore alpha layer rule...

PR #1106, opened 2026-06-29
feat(gestures): 1:1 touchpad tag scrubbing

ernestoCruz05:
1to1 gestures for when changing tags, i also worked a bit on something like this for overview but i found it to be too troublesome and not a good workflow

https://github.com/user-attachments/assets/bd096460-6def-48a4-ada5-4fa5fbe6fc87

References #937

### DIFF
--- a/assets/config.conf
+++ b/assets/config.conf
@@ -120,6 +120,7 @@ swipe_min_threshold=1
 gesturebind=none,horizontal,4,tagscrub
 gesture_swipe_distance=500
 gesture_commit_ratio=0.5
+gesture_axis_lock=16
 
 # mouse
 # need relogin to make it apply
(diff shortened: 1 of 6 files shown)


DreamMaoMao:
This architecture seems a bit complicated. I have other design ideas,
 but I haven't had the time to implement them yet.
 
 
image

The sliding distance control factor can be directly introduced, and then the current time can be continuously updated to the corresponding proportional time. In this way, when the hand is released, it can continue to move according to the current proportion. Only one auxiliary control variable and a small amount of code need to be introduced

ernestoCruz05:
Thanks for the feedback, especially the client_animation_next_tick feedback, I will go back to the drawing board and make a better implemmentation, or at least, alter the current implemmentation

DreamMaoMao:
finish in latest commit.

PR #1123, opened 2026-07-05
feat(nix): add hjem-module based on home-manager module

Nerglej:
Hi there!

I'm migrating my config to [hjem](https://github.com/feel-co/hjem) from home-manager. Since they're very much alike module-wise, I made a common config helper that configures both the home-manager and hjem modules. I get identical output with home-manager before and after my changes, and also with hjem now.

I made some decisions that I would like opinions on:

- I chose to call the hjem option `desktops.mango`, since there's no established hjem system except for how [hjem-rum](https://github.com/snugnug/hjem-rum) does it.
- Added two outputs, both `mango.hjemModules.mango` and `mango.hjemModules.default`, since I've seen `mango.hjemModules.default` more often in the hjem ecosystem, e.g. [hjem-rum](https://github.com/snugnug/hjem-rum/blob/main/flake.nix) and [noctalia](https://github.com/noctalia-dev/noctalia)
- I decided to let the warning about the legacy text based config stay for home-manager only, since this hjem-module has never had the text based config.

Have a nice day :D

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,53 +8,60 @@
     };
   };
 
-  outputs = {
-    self,
-    flake-parts,
-    ...
-  } @ inputs:
-    flake-parts.lib.mkFlake {inherit inputs;} {
+  outputs =
+    {
+      self,
+      flake-parts,
+      ...
+    }@inputs:
+    flake-parts.lib.mkFlake { inherit inputs; } {
       imports = [
         inputs.flake-parts.flakeModules.easyOverlay
       ];
 
       flake = {
         hmModules.mango = import ./nix/hm-modules.nix self;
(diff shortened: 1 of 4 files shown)


Nerglej:
I just realized that I formatted flake.nix. Do we have any nix styling guidelines?

rysieko:
for formatter there's outputed  `formatter = pkgs.alejandra;` in the flake, set your LSP to format using `nix fmt`

Nerglej:
Since, I've made this pull request, the formatter for nix files have changes to nixfmt, which is what I used initially xD. I've just rebased the commits since this pr, and it works great.

PR #1139, opened 2026-07-08
fix: printstatus when selmon changes on cursor motion

WeraPea:


duarm:
related? https://github.com/mangowm/mango/issues/1141

PR #1140, opened 2026-07-08
opt: focus selected client on monocle/fullscreen monitors

WeraPea:
When using focusdir with focus_cross_monitor=1, and focusing to a monitor that is monocle or fullscreen, it now focuses the monitors selected client, instead of one that is possibly invisible to the user.


DreamMaoMao:
refer

a7b1e0b50116d0ca6bf5b1b7861ce73d80529bf8

PR #1163, opened 2026-07-11
feat: hdr:2 monitor rule value — read HDR mastering data from EDID

m8l8th814n-eng:
## What

Adds a new value for the `hdr` monitor rule: `hdr:2`.

`hdr:1` sends an image description with BT2020 primaries and PQ transfer function, but leaves the mastering luminance, max CLL/FALL, and mastering primaries unset — so tone mapping runs against generic defaults rather than what the panel can actually do.

`hdr:2` does everything `hdr:1` does, and additionally parses the output's EDID (via libdisplay-info) to fill the image description with the panel's real HDR static metadata:

- mastering display min/max luminance
- max content light level (CLL) and max frame-average light level (FALL)
- mastering display color primaries and white point

This is equivalent to Hyprland's `cm = hdredid` mode. On panels whose peak luminance differs a lot from the defaults (e.g. laptop OLEDs), this makes highlights and overall brightness track the display's actual capabilities.

## Implementation

- `output_fill_edid_hdr_caps()` in `src/ext-protocol/hdr.h` reads `/sys/class/drm/card*-<name>/edid` and extracts the metadata with libdisplay-info. If the EDID can't be read or has no HDR metadata block, it silently falls back to plain `hdr:1` behavior.
- `Monitor.hdr_enable` widened from `bool` to `int32_t`; config clamp changed from `0..1` to `0..2`. `hdr:1` behavior is unchanged.
- New dependency: `libdisplay-info` (added to `meson.build` and `nix/default.nix`).
- Docs table in `docs/configuration/monitors.md` updated.

## Testing

Running daily on a Samsung ATNA56YX03 OLED (eDP-1, max 617 nits per EDID) with `monitorrule=...,hdr:2`. `drm_info` shows Colorspace=BT2020_RGB and the HDR_OUTPUT_METADATA blob populated; the debug log prints the parsed EDID values (max 617, avg 400, min 0.0050). Builds clean against wlroots 0.20 + scenefx 0.5.


m8l8th814n-eng (commented):
Still rocking :)

PR #1183, opened 2026-07-18
changed spawn_on_empty behaviour

AgrimV:
this is related to [this](https://github.com/mangowm/mango/discussions/1170) discussion

fixed(?) view switching on `spawn_on_empty` to be better handled so as to not impede other dispatchers
the discussion did not gain much traction, so i am implementing what i think should be the default behavior, this was the simplest solution i could think of, sorry if i did something incorrectly

tested on fedora 44


AgrimV:
#1377 solves this issue, and is more likely to be merged

Issue #1184, opened 2026-07-18
dragging a window and switching to another tag causes dropcolor to be displayed but no window tag switching.


azrtareef:
mango version: mango 0.15.3(89264e7) (wl-only branch)
wlroots version: 0.20

As the title of this issue suggests, when i drag a specific window and use the keybinding to switch to another tag, the configured dropcolor does get displayed but the movement doesn't occur. If this is the expected behaviour, it would honestly be better if the dropcolor wasnt being shown in the first place. Following is a method to reproduce the same thing:

*Make sure a visible drop color is configured*
1. Spawn a terminal in tag 1 and tag 2.
2. Return to tag 1 and select the window to be moved and use your configured keybinding to switch to tag 2
3. Move the window around and observe the dropcolor being displayed.

Here is a screenrecording depicting the issue:

https://github.com/user-attachments/assets/edc7f244-2db2-4c09-b933-ae583288fd95

azrtareef:
Thanks to @darealhoks for creating a PR and to @DreamMaoMao for merging it.

However, after the latest commit, ive noticed that while the original issue has been fixed (the issue related to displaying the dropcolor on another tag but not moving the window), doing the same thing on a tag with no windows does not display the preview created by the `drag_tile_small` option. I would appreciate if this feature can be added as it provides more feedback to the user regarding window switching. Here is a screen-recording concering the feature in question for more clarification:

https://github.com/user-attachments/assets/9e70163c-b39b-4f04-8ff4-1efffe35f762

PR #1191, opened 2026-07-20
fix: dragged window hidden after switching tag mid-drag

darealhoks:
Follow-up to #1184 ([comment](https://github.com/mangowm/mango/issues/1184#issuecomment-5020241345)): dragging a window to an empty tag shows no drag_tile_small preview, the window just vanishes until you drop it.

## Cause
The grabbed client keeps its old tags mid-drag, so arrange() puts it in the hidden branch on a view switch. It only stayed visible when the tag-out animation happened to keep its node enabled (animations on, and only when leaving its home tag), which is why it works on some tags and not others.

## Fix
Skip set_arrange_hidden for grabc. Its position is already driven by the cursor every frame, and it gets retagged and re-tiled on release as before.

Tested on Void Linux, wlroots 0.20.2 / scenefx 0.5: preview stays under the cursor on empty and populated tags, animations on or off, any number of tag switches.

### DIFF
--- a/src/layout/arrange.h
+++ b/src/layout/arrange.h
@@ -1204,7 +1204,8 @@ void pre_caculate_before_arrange(Monitor *m, bool want_animation,
 				if (!only_caculate)
 					set_arrange_visible(m, c, want_animation);
 			} else {
-				if (!only_caculate)
+				/* keep the dragged client visible across view switches */
+				if (!only_caculate && c != grabc)
 					set_arrange_hidden(m, c, want_animation);
 			}
 		}


DreamMaoMao:
There is an ambiguity here. The preview mentioned in the issue is not visible, not the window

darealhoks:
You're right, there are two different things in the video and i accidentally fixed the wrong one, but i also feel like my fix makes sense. On the video, the small window does also disapear when switching with a dragged window to another tag. The preview the issue is actually complaining about is a new feature, should i add that to this or retitle it to make the scope clearer and make another pull request for the other? Or would you rather not add it at all?

DreamMaoMao:
> You're right, there are two different things in the video and i accidentally fixed the wrong one, but i also feel like my fix makes sense. On the video, the small window does also disapear when switching with a dragged window to another tag. The preview the issue is actually complaining about is a new feature, should i add that to this or retitle it to make the scope clearer and make another pull request for the other? Or would you rather not add it at all?

10302fd5064325cb58682bbb126622a5fb103876

Wasn't this issue fixed in the previous comomit?

darealhoks:
> > You're right, there are two different things in the video and i accidentally fixed the wrong one, but i also feel like my fix makes sense. On the video, the small window does also disapear when switching with a dragged window to another tag. The preview the issue is actually complaining about is a new feature, should i add that to this or retitle it to make the scope clearer and make another pull request for the other? Or would you rather not add it at all?
> 
> [10302fd](https://github.com/mangowm/mango/commit/10302fd5064325cb58682bbb126622a5fb103876)
> 
> Wasn't this issue fixed in the previous comomit?

No, that was only the logic of it, the previous commit allowed you to drop a dragged window on another tag, but the small window when you drag is now only visible across tags if:
1. Leaving its original tag, if you drag a window and switch from tag 1 to tag 2, the small window is still visible, but if you then switch to any other tag it disapears.
2. Animations are enabled, with animations disabled, the small window disapears immidiately as you switch tags

This PR fixes it: you are able to see the small window regardless of how many tags you switch between and whether your animations are turned on or not.

azrtareef:
@darealhoks thanks you for creating a pr regarding my issue. I have tested the PR and this is exactly what I was referring to, having the preview window still being visible even while switching the tags via a binding for more feedback. I can confirm that this pr completely supports that.

PR #1200, opened 2026-07-22
fix: reapply VRR on physical monitor reconnect

rockerBOO:
The monitor VRR was not getting set after turning off and turning on the monitor. 

The VRR attempts to apply when the monitor is focused. There is a retry mechanism to attempt to set it in a loop to let the monitor settle. 

The following code is to handle when the output commit fails it will remain in "enabling" mode. But this pattern is kinda sloppy though and duplicated a few times. 

```c
	    if (mango_output_commit(m))
			updatemons(NULL, NULL);
		else
			m->is_vrr_enabling = true;
```

This issue is likely similarly happening for HDR but that has a larger number of changes and harder to check because tooling doesn't support checking HDR support quite yet. I have a branch of `wlr-randr` to support looking at HDR support and a monitor that supports HDR so I can look into this if an approach here is acceptable. 

Willing to make changes and feedback if the HDR changes would be desired.

### DIFF
--- a/src/action/monitor.h
+++ b/src/action/monitor.h
@@ -56,7 +56,8 @@ bool mango_output_commit(Monitor *m) {
 		wlr_output_state_finish(&m->pending);
 		wlr_output_state_init(&m->pending);
 	} else {
-		wlr_log(WLR_ERROR, "Failed to commit frame");
+		wlr_log(WLR_ERROR, "Failed to commit frame for output %s",
+				m->wlr_output->name);
 	}
 	return committed;
 }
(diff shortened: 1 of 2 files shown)


DreamMaoMao:
why not wlr-randr --output eDP-1 --on --adaptive-sync enabled use this?

rockerBOO:
Every time I turn my monitor on again I would need to re-run the wlr-randr command to enable it. We already are setting the VRR in our Mango config so makes it more consistent than to silently drop VRR when the monitor power refreshes. I had a udev command to detect the monitor coming on and running wlr-randr but it seemed like a workaround to getting Mango to re-enable VRR.  

Edit: 

Also could remove the "retry" parts as it will try to apply the VRR when the cursor enters the screen so that trade off is probably fine.

PR #1206, opened 2026-07-22
fix: don't pass key release event when it's handled by a keybinding

### LINKED ISSUE #348: [Bug] `SUPER+SPACE` keybind passed to client even though it's bound to keybind
Hi

I have `bind=SUPER,space,spawn,~/.config/scripts/toggle-rofi.sh`
When playing youtube and launching rofi, SUPER+SPACE launches rofi correctly, but youtube also registers the space and pauses, which is weird to me. Other keybinds like `SUPER+k` (k should also play/pause youtube) don't have this issue, and youtube keeps playing.

Lubsch:
Also fixes the opposite case: Don't pass key press event when it's handled by an on-release keybinding.

This makes the keybinding function always return true (meaning "handled") on both press and release for mappings even when it didn't actually execute the function because it's mapped to the opposite event type. This information is used by the keypress function to decide if it should propagate the event to the client.

Fixes https://github.com/mangowm/mango/issues/348

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -4326,30 +4326,21 @@ keybinding(uint32_t state, bool locked, uint32_t mods, xkb_keysym_t sym,
 	int32_t ji;
 	int32_t isbreak = 0;
 
+	if (state != WL_KEYBOARD_KEY_STATE_PRESSED &&
+		state != WL_KEYBOARD_KEY_STATE_RELEASED) {
+		return 0;
+	}
+
 	if (is_keyboard_shortcut_inhibitor(seat->keyboard_state.focused_surface)) {
-		return false;
+		return 0;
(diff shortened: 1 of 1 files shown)


Lubsch:
See also: https://codeberg.org/dwl/dwl/pulls/1108#issuecomment-5063106

DreamMaoMao:
I'm going to roll back this. In some cases, it will cause the repeated key to be triggered all the time

Lubsch:
> I'm going to roll back this. In some cases, it will cause the repeated key to be triggered all the time

I understand! I've also encountered that using this for a few hours. Sorry that I didn't notice it earlier. I might come back with another patch. Thanks for being open to the contribution

PR #1209, opened 2026-07-23
fix: keybinding release propagation (attempt 2)

Lubsch:
See also https://github.com/mangowm/mango/pull/1206 and https://github.com/mangowm/mango/pull/1206#issuecomment-5053777350

With the previous commit, there was a bug causing key repeats to not stop after releasing the button when a keybinding was pressed during key hold.

Steps to replicate:

- Assume SUPER+j is mapped to something on press
- hold j -> j is repeated (expected)
- hold SUPER -> j is still repeated, no binding executed (expected)
- release j -> j is still repeated (WRONG)
- release SUPER -> j is still repeated (WRONG)

You then must press j again to end the key repeat

This commit adds a 256 byte bool array `intercepted` to `KeyboardGroup` which stores whether a key's press was intercepted and shouldn't be passed to clients.

I don't think this is the most elegant solution but something similar is done by other compositors
like sway: https://github.com/swaywm/sway/blob/6959a78a8f0d52f79ad7465135b3295307a5146a/include/sway/input/keyboard.h#L29
or river: https://codeberg.org/river/river-classic/src/commit/b1af960e1bea8506c25179775c18f902a5cb8c87/river/Keyboard.zig#L41

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -480,6 +480,8 @@ typedef struct {
 	const xkb_keysym_t *keysyms; /* invalid if nsyms == 0 */
 	uint32_t mods;				 /* invalid if nsyms == 0 */
 	uint32_t keycode;
+	bool intercepted[256]; // press was consumed by a binding, indexed by
+						   // event->keycode
 	struct wl_event_source *key_repeat_source;
 
 	struct wl_listener modifiers;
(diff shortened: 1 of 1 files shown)


Lubsch:
I've just noticed that this now propagates release event for release keybinds when it shouldn't. I'Lll undraft this when I've properly tested it.

PR #1213, opened 2026-07-23
Fix/restore minimized

### LINKED ISSUE #1211: Incorrect focusing when calling restore_minimized with multi monitor setups
## Info
mango version(mango -v): mango 0.15.4(03caa97)
wlroots version: wlroots 0.20.2-1

There is a logic bug with the focusing of windows unminimized in multi monitor setups.

### What happens
- Minimize window on monitor 1, tag 1
- On monitor 1, go to tag X
- Swap focus to monitor 2, any tag
- Call restore_minimized from monitor 2, any tag
- Monitor 2 will focus tag 1
- Unminimized client in monitor 1, tag 1 is focused, however it is not visible as monitor 1 is still focused on tag X

### Example video:
Left is laptop screen
Right is ultrawide monitor

https://github.com/user-attachments/assets/dd0fdc8a-dcd4-4c52-a33b-f7296fbf8097

lefrogfrog:
Fixes #1211

`restore_minimized` calls `show_hide_client`, which called `tag_client()` which internally called `view()`, always operating on `selmon`. When restoring a minimized client from a different monitor than the client belongs to, the wrong monitor's tag
was switched, leaving the restored client invisible on its own monitor.

Fix by inlining the tag_client logic in `show_hide_client` and calling `view_in_mon` directly on the client's monitor (`c->mon`) instead of `selmon`, so the correct monitor switches to the restored client's tag.

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -6106,6 +6106,16 @@ void show_hide_client(Client *c) {
 
 	if (!c->is_in_scratchpad) {
 		tag_client(&(Arg){.ui = target}, c);
+		c->tags = target;
+		c->istagswitching = 1;
+		Client *fc = NULL;
+		wl_list_for_each(fc, &clients, link) {
+			if (fc && fc != c && c->tags & fc->tags && ISFULLSCREEN(fc) &&
+				!c->isfloating) {
+				clear_fullscreen_flag(fc);
+			}
+		}
+		view_in_mon(&(Arg){.ui = target}, true, c->mon, false);
 	} else {
 		c->tags = c->oldtags;
 		arrange(c->mon, false, false);
--- a/src/mango.c
+++ b/src/mango.c
@@ -6105,7 +6105,6 @@ void show_hide_client(Client *c) {
 	target = get_tags_first_tag(c->oldtags);
 
 	if (!c->is_in_scratchpad) {
-		tag_client(&(Arg){.ui = target}, c);
 		c->tags = target;
 		c->istagswitching = 1;
 		Client *fc = NULL;


DreamMaoMao:
refer:

de0200c49000200114e0e7b98dbf477ea8f8d4df

lefrogfrog:
@DreamMaoMao 

Thanks for adding the new 0/1 flag! I just tested it with restore_minimized, 1, but the multi-monitor bug from #1211 is still happening.

If I trigger the shortcut from Monitor 2, the system still incorrectly switches Monitor 2 to Tag 1, while leaving the actual unminimized window completely hidden over on Monitor 1.

Could we reopen issue #1211? I'd be happy to update my PR to work alongside the new 0/1 logic if you'd like.

DreamMaoMao:
> @DreamMaoMao
> 
> Thanks for adding the new 0/1 flag! I just tested it with restore_minimized, 1, but the multi-monitor bug from #1211 is still happening.
> 
> If I trigger the shortcut from Monitor 2, the system still incorrectly switches Monitor 2 to Tag 1, while leaving the actual unminimized window completely hidden over on Monitor 1.
> 
> Could we reopen issue #1211? I'd be happy to update my PR to work alongside the new 0/1 logic if you'd like.

To be honest, I didn't understand what your show.

lefrogfrog:
@DreamMaoMao  Yea the video was a bit confusing specially with how squished everything was my bad. 

Let me break down the exact issue step-by-step.

The bug happens when you try to un-minimize a window while your focus is on a different monitor than the monitor in which you originally minimized the window itself.

# Step-by-Step Reproduction:

1. You have two monitors: Monitor A and Monitor B.
2. Open a window on Monitor A on Tag 1, and minimize it.
3. Still on Monitor A, switch to a different workspace (e.g., Tag 2).
4. Move your mouse/focus over to Monitor B.
5. While focused on Monitor B, trigger `restore_minimized`.

# The Bug (What goes wrong):
Because you triggered the shortcut from Monitor B, MangoWM incorrectly forces Monitor B to switch to Tag 1.
Meanwhile, the window un-minimizes on Monitor A, but because Monitor A is still looking at Tag 2, the window is completely invisible. The window technically gets keyboard focus, but you can't see it.

# Expected Behavior:
When you trigger the un-minimize shortcut, MangoWM should look at the window's native monitor (Monitor A). It should switch Monitor A back to Tag 1 so the window is actually visible, and leave Monitor B alone.

Essentially, the function seems to be applying the tag-switch logic to the currently selected monitor (selmon) instead of the monitor where the client actually lives (c->mon).


Let me know if the issue is unclear! 

My PR corrects this specific monitor-targeting logic.

DreamMaoMao:
you can check the latest commit again.

lefrogfrog:
@DreamMaoMao the window now correctly gets focused on monitor A, however monitor B still  incorrectly switches focus.

The call chain looks like this: `show_hide_client` ->  `tag_client` -> `view` -> `view_in_mon` which uses `selmon`, because `selmon` has not yet been updated, monitor B to incorrectly switches focus.

A simple one line fix could be to set `selmon=c->mon` in `restore_minimized` before the call to `show_hide_client` or somewhere along the chain before `view_in_mon`.

DreamMaoMao:
should work now.

lefrogfrog:
@DreamMaoMao now it works perfectly, thanks! 🫶

PR #1231, opened 2026-07-27
feat: add dwindle_toggle_current_split dispatch

lefrogfrog:
Toggles the parent split node's split_h on the focused window and reflows immediately, unlike dwindle_toggle_split_direction which only affects future window placement.

3 files changed:
- bind_declare.h — declaration
- bind_define.h — finds leaf → parent, flips split_h, sets split_locked=true, calls arrange()
- parse_config.h — register "dwindle_toggle_current_split" in parse_func_name()

Keybind example: bind = Super, S, dwindle_toggle_current_split

### DIFF
--- a/src/dispatch/bind_define.h
+++ b/src/dispatch/bind_define.h
@@ -2185,7 +2185,6 @@ void dwindle_set_split_direction(Client *c, bool istoggle, bool horizontal) {
 		return;
 
 	if (istoggle) {
-		// NOTE: we do this
 		leaf->custom_leaf_split_h = !leaf->custom_leaf_split_h;
 	} else if (horizontal) {
 		leaf->custom_leaf_split_h = true;
--- a/src/dispatch/bind_define.h
+++ b/src/dispatch/bind_define.h
@@ -2238,7 +2238,7 @@ void dwindle_toggle_current_split(const Arg *arg) {
 	if (layout->id != DWINDLE)
 		return;
 
-	DwindleNode *root = c->mon->pertag->dwindle_root[selmon->pertag->curtag];
+	DwindleNode *root = c->mon->pertag->dwindle_root[c->mon->pertag->curtag];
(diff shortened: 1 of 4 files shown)


src/dispatch/bind_define.h
```diff
@@ -2227,6 +2227,29 @@ void dwindle_split_vertical(const Arg *arg) {
...
+		return;
+	Client *c = arg->tc ? arg->tc : selmon->sel;
+	if (!c || !c->mon || c->isfloating)
+		return;
+
+	const Layout *layout = c->mon->pertag->ltidxs[c->mon->pertag->curtag];
+	if (layout->id != DWINDLE)
+		return;
+
+	DwindleNode *root = selmon->pertag->dwindle_root[selmon->pertag->curtag];
```
DreamMaoMao: shouldn't use selmon , should be c->mon, because the c may be arg->tc not selmon->sel

lefrogfrog:
@DreamMaoMao `dwindle_toggle_current_split` now uses c->mon, and I added it to the docs

src/dispatch/bind_define.h
```diff
@@ -2227,6 +2227,29 @@ void dwindle_split_vertical(const Arg *arg) {
...
+		return;
+	Client *c = arg->tc ? arg->tc : selmon->sel;
+	if (!c || !c->mon || c->isfloating)
+		return;
+
+	const Layout *layout = c->mon->pertag->ltidxs[c->mon->pertag->curtag];
+	if (layout->id != DWINDLE)
+		return;
+
+	DwindleNode *root = c->mon->pertag->dwindle_root[selmon->pertag->curtag];
```
DreamMaoMao: still selmon in the laste pointer

lefrogfrog: All c->mon now 👍

PR #1234, opened 2026-07-28
fix(nix): disable upstream nixpkgs programs.mango module

### LINKED ISSUE #1233: nix: nixpkgs update breaks the mango nixos module
## Checklist

## Info

Someone in nixpkgs updated the mangowc module, to have the prefix `programs.mango.*` (before it was using the `programs.mangowc.*`). This conflicts with the official NixOS module from mango, and when trying to evaluate a system configuration using the latest nixpkgs unstable channel coupled with the official mango nixos modules results into a conflict and does not evaluate at all.

yvnth:
## Disable upstream nixpkgs `programs.mango` module

Fixes #1233

nixpkgs recently added its own `programs.mango` module (previously
`programs.mangowc`), which collides with this flake's
`nixosModules.mango` since both declare `programs.mango.*`. Evaluating
a system with both imported fails:

```
error: The option `programs.mango.enable' ... is already declared in ...
```

First commit adds `disabledModules = [ "programs/wayland/mango.nix" ]`
to `nix/nixos-modules.nix` so nixpkgs's copy is disabled automatically
whenever `nixosModules.mango` is imported.

Second commit switches the flake `formatter` from `alejandra` to the
official nixpkgs formatter, `nixfmt`, and reformats the touched files
accordingly.

Tested: eval and full rebuild succeed, `mango -v` runs fine post-build,
`nix flake check --all-systems` passes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,12 +8,13 @@
     };
   };
 
-  outputs = {
-    self,
-    flake-parts,
-    ...
-  } @ inputs:
-    flake-parts.lib.mkFlake {inherit inputs;} {
+  outputs =
+    {
+      self,
+      flake-parts,
+      ...
+    }@inputs:
+    flake-parts.lib.mkFlake { inherit inputs; } {
       imports = [
         inputs.flake-parts.flakeModules.easyOverlay
       ];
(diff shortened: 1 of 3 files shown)


DreamMaoMao:
Since it has been added upstream, can we remove it here?

yvnth:
I'd personally be against removing it. If we drop it here, everyone using the flake method would have to switch over to the nixpkgs method themselves. Since I'm the only active `mango` maintainer in nixpkgs right now (the co-maintainer isn't active), even a version bump can take up to 10 days to land. Some people may prefer keeping the flake module so they can get the latest version whenever they want, without waiting on that.

PR #1250, opened 2026-08-06
fix: activate pointer constraints on layer-shell surfaces

### LINKED ISSUE #1080: lan-mouse can’t keep the mouse on client’s side
## Info

mango version: 0.14.4(release)
wlroots version: idk

## Description

it’s the issue that https://github.com/mangowm/mango/pull/865 is supposed to solve but it is not. I’m installing mango through NixOS flake so I think it is on main branch. I might be wrong but the merge was in April and 14.4 is 5d old so it should be in.

Note that the client was using Niri at the time of testing, can that be an issue?

bchow000:
Fixes #1080

## Problem

#865 fixed the `motionnotify` half of this, but nothing ever sets
`active_constraint` for a layer surface, so that code is never reached.

`active_constraint` is only set in two places, and both look for a toplevel:
`createpointerconstraint` compares against `client_surface(selmon->sel)`, and
`focusclient` searches the constraint list for `client_surface(c)`. A layer
surface is never `selmon->sel`, so `lock_pointer` from one is accepted but never
activated — `send_activated` is never sent, the cursor keeps moving, leaves the
surface, and the client gets a `leave`.

lan-mouse locks the pointer on a 1px layer surface, so capture only holds while
the cursor sits exactly on the border strip and drops as soon as it moves off.

## Fix

- `createpointerconstraint` — also match `seat->pointer_state.focused_surface`,
  for clients that lock from their pointer enter handler.
- `pointerfocus` — activate a constraint on the focused surface when it has no
  client (`xytonode` returns layer surfaces through `pl`, never `pc`).
  Activate-only, so it won't grab the pointer for an unfocused window with
  `sloppyfocus=0`.

Tested on 0.15.5, CachyOS, wlroots 0.20, with lan-mouse's layer-shell capture
backend. Confirmed via `WAYLAND_DEBUG=client` that `zwp_locked_pointer_v1.locked`
now reaches the client; before the change it never arrived.

### DIFF
--- a/src/mango.c
+++ b/src/mango.c
@@ -3902,6 +3902,13 @@ void createpointerconstraint(struct wl_listener *listener, void *data) {
 	LISTEN(&pointer_constraint->constraint->events.destroy,
 		   &pointer_constraint->destroy, destroypointerconstraint);
 
+	// layer surfaces are never selmon->sel, so match pointer focus too
+	if (seat->pointer_state.focused_surface ==
+		pointer_constraint->constraint->surface) {
+		cursorconstrain(pointer_constraint->constraint);
+		return;
+	}
+
 	if (!selmon || !selmon->sel)
 		return;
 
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
how about the latest change?

bchow000:
> how about the latest change?

Seems to be working well, no issues so far.

DreamMaoMao:
> > how about the latest change?
> 
> Seems to be working well, no issues so far.

image

how about this?

bchow000:
> > > how about the latest change?
> > 
> > 
> > Seems to be working well, no issues so far.
> 
> image
> how about this?

The mouse moving to specific areas and then getting constrained is exactly what lan-mouse does.  It uses an invisible overlay to capture the mouse position, and when the user moves the mouse to a side of the screen that’s configured relative to the position of a second monitor attached to a second computer. It becomes constrained on the primary computer and moves about on the secondary system until it comes back to coordinates that represent the primary computer or the client disconnects.  At which point the mouse is released on the primary system again.

PR #1259, opened 2026-08-07
hdr: mastering display metadata, runtime toggle, and damage on image description change

ezvk:
Three related HDR changes. Everything below was measured on hardware — a Samsung ATNA40CU05-0 OLED (internal, amdgpu) and an RTK ZEUSLAP TYPEC over USB-C — under wlroots 0.20 with the Vulkan renderer. Single commit, `format.sh` applied.

Targets `wl-only` since that is where HDR lives.

## 1. Mastering display metadata

`output_enable_hdr()` fills only `.primaries` and `.transfer_function` of `wlr_output_image_description`. The other four fields stay zero with a designated initializer, so the panel receives an HDR infoframe with 0.0 primaries and zero luminance in the mastering block, and has to guess how to tone-map.

wlroots does not expose the EDID luminances — `wlr_output` only carries `supported_primaries` and `supported_transfer_functions` — so the values come from `monitorrule` rather than a second EDID parse:

| key | meaning |
| :--- | :--- |
| `hdr_min_lum` | mastering minimum luminance, cd/m² |
| `hdr_max_lum` | mastering peak, also sent as `max_cll` |
| `hdr_max_avg_lum` | `max_fall` |
| `hdr_force` | enable HDR even when the EDID does not advertise BT.2020/PQ |

```ini
monitorrule=name:eDP-1,...,hdr:1,hdr_max_lum:616.884,hdr_max_avg_lum:400
```

Leaving them unset leaves the fields at zero, so **existing configs behave exactly as before**.

`hdr_force` is for panels that declare HDR only inside a **DisplayID 2.0** extension, with the CTA-861 blocks nested in a `0x81` container. That is legal EDID 1.4, but libdisplay-info`s CTA path never descends into it, so `supported_primaries` comes back empty and `hdr:1` is silently dropped on hardware that drives PQ fine. It skips those two EDID-derived checks only — the `output_color_transform` check stays, since that is a real renderer capability and no config key should be able to fake it.

## 2. togglehdr

```
togglehdr[,on|off|toggle][,<monitor name>|all]
```

The runtime equivalent of sway`s `output <name> hdr on|off|toggle`. HDR could only be set from `monitorrule`, so changing it meant a config reload that also re-applies mode, scale, position and transform on every output.

With `all`, toggle takes **one** decision for every output — if anything is on, everything goes off — instead of flipping each monitor against its own state, which would leave a multi-monitor desk half on and half off and swap the halves on the next press. Outputs that cannot do HDR are skipped by the existing `output_supports_hdr()` check, which already runs before any state is mutated.

Two details worth flagging for review: the commit is forced, because `mango_scene_output_commit()` returns early when `wlr_scene_output_needs_frame()` is false and a still screen would otherwise swallow the change; and `wlr_output_state.allow_reconfiguration` must be set, or wlroots refuses the commit outright.

## 3. Damage the whole output when the image description changes

wlroots calls `scene_output_damage_whole()` for a geometry, transform, scale or rendered-gamma-LUT change — but **not** for an image description change (same call sites in 0.20.2 and master). Switching between PQ/BT.2020 and sRGB changes how every pixel has to be encoded, so anything left undamaged keeps the luminance mapping it had when last drawn. On a static desktop that is most of the screen, and the result is flat rectangles at the wrong brightness that survive further toggles.

Ruled out as a gamma problem by control: with the gamma LUT at 1.0, the daemon killed and no `zwlr_gamma_control_v1` in existence at all, the artefact still appeared.

Filling the damage ring is enough and uses only public API — `wlr_scene.c` derives the re-rendered region from `wlr_damage_ring_rotate_buffer()`. The scene also unions into `pending_commit_damage`, but that is private and only feeds the damage rectangle of the DRM commit, and a reconfiguration scans out the whole frame regardless.

## Measured result

Two panels, each fed from its own EDID:

```
eDP-1  Colorspace 9 | eotf 2 | bpc 10 | mastering 616 | max_cll 616 | max_fall 400
DP-2   Colorspace 9 | eotf 2 |          mastering 417 | max_cll 417 | max_fall 417
```

Both were `0/0/0` before. The second one also works across a multi-GPU boundary — rendered on the AMD iGPU, scanned out on a connector belonging to the discrete NVIDIA GPU.

### DIFF
--- a/docs/configuration/monitors.md
+++ b/docs/configuration/monitors.md
@@ -31,6 +31,10 @@ monitorrule=name:Values,Parameter:Values,Parameter:Values
 | `scale` | float | 0.01-100.0 | Monitor scale |
 | `vrr` | integer | 0, 1 | Enable variable refresh rate |
 | `hdr` | integer | 0, 1 | Enable hdr support |
+| `hdr_min_lum` | float | 0.0-10000.0 | Mastering display minimum luminance, cd/m² (0 = unset) |
+| `hdr_max_lum` | float | 0.0-10000.0 | Mastering display peak luminance, also sent as max_cll, cd/m² (0 = unset) |
+| `hdr_max_avg_lum` | float | 0.0-10000.0 | Max frame-average light level (max_fall), cd/m² (0 = unset) |
+| `hdr_force` | integer | 0, 1 | Enable HDR even when the EDID does not advertise BT.2020/PQ |
 | `rr` | integer | 0-7 | Monitor transform |
 | `custom` | integer | 0, 1 | Enable custom mode (not supported on all displays — may cause black screen) |
 | `disable` | integer | 0, 1 | Disable the monitor |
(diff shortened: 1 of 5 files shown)


DreamMaoMao:
There are a large number of irrelevant test comments in the submission and documentation

ezvk:
cleaning

ezvk:
You are right — thanks for the review. I had left my debugging notes in the code and the docs.

Cleaned up and force-pushed:

- comment lines in the diff: **118 → 37**, total insertions **365 → 243**
- removed all the test/measurement notes: dates, panel model names, step-by-step reproductions, references to my own setup
- kept only the short non-obvious rationale: why `allow_reconfiguration` is required, why the commit is forced, why both flags are snapshotted before a possibly failing commit, and why the damage ring is filled
- documentation trimmed the same way — the three sections are now just what a user needs to configure the keys, without the war stories
- commit message shortened as well
- `format.sh` re-applied, builds clean on `wl-only`

Happy to trim further, or to split the three parts into separate PRs if you would rather review them independently.

PR #1267, opened 2026-08-10
feat: Add way to control all outer gap individualy

Wateir:
You can define a different gap on the right and on the left, same for top and bottom. Intorudce 4 new settings

| Setting | Default | Description |
| :--- | :--- | :--- |
| `gappol` | `10` | Outer gap, left edge. |
| `gappor` | `10` | Outer gap, right edge. |
| `gappot` | `10` | Outer gap, top edge. |
| `gappob` | `10` | Outer gap, bottom edge. |

For backward compatibility, `gappoh` sets both `gappol` and `gappor`, and `gappov` sets both `gappot` and `gappob`. If you mix legacy and directional keys, the last one in the file wins.

i have run `./format.sh`

I have compile and test them, they work fine, and the backward compatibility too work

### DIFF
--- a/docs/visuals/theming.md
+++ b/docs/visuals/theming.md
@@ -12,8 +12,12 @@ Control the sizing of window borders and gaps.
 | `borderpx` | `4` | Border width in pixels. |
 | `gappih` | `5` | Horizontal inner gap (between windows). |
 | `gappiv` | `5` | Vertical inner gap. |
-| `gappoh` | `10` | Horizontal outer gap (between windows and screen edges). |
-| `gappov` | `10` | Vertical outer gap. |
+| `gappot` | `10` | Outer gap, top edge. |
+| `gappor` | `10` | Outer gap, right edge. |
+| `gappob` | `10` | Outer gap, bottom edge. |
+| `gappol` | `10` | Outer gap, left edge. |
+
+> **Tip:** For backward compatibility, `gappoh` sets both `gappol` and `gappor`, and `gappov` sets both `gappot` and `gappob`. If you mix legacy and directional keys, the last one in the file wins.
 
 ## Colors
 
--- a/src/layout/horizontal.h
+++ b/src/layout/horizontal.h
@@ -16,8 +16,10 @@ void tile(Monitor *m) {
 
 	int32_t cur_gappiv = enablegaps ? m->gappiv : 0;
 	int32_t cur_gappih = enablegaps ? m->gappih : 0;
-	int32_t cur_gappov = enablegaps ? m->gappov : 0;
-	int32_t cur_gappoh = enablegaps ? m->gappoh : 0;
+	int32_t cur_gappol = enablegaps ? m->gappol : 0;
+	int32_t cur_gappor = enablegaps ? m->gappor : 0;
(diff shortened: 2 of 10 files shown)


Wateir:
@ernestoCruz05 possible to review this if you have some time ?

docs/visuals/theming.md
```diff
@@ -12,8 +12,12 @@ Control the sizing of window borders and gaps.
...
 | `gappih` | `5` | Horizontal inner gap (between windows). |
 | `gappiv` | `5` | Vertical inner gap. |
-| `gappoh` | `10` | Horizontal outer gap (between windows and screen edges). |
-| `gappov` | `10` | Vertical outer gap. |
+| `gappol` | `10` | Outer gap, left edge. |
+| `gappor` | `10` | Outer gap, right edge. |
+| `gappot` | `10` | Outer gap, top edge. |
+| `gappob` | `10` | Outer gap, bottom edge. |
+
+> **Tip:** For backward compatibility, `gappoh` sets both `gappol` and `gappor`, and `gappov` sets both `gappot` and `gappob`. If you mix legacy and directional keys, the last one in the file wins.
```
ernestoCruz05, line 20: I would reconsider this maybe, instead of going left right top bottom I would go top right bottom left as thats a more "human" way to read, it's just UX, I think the Tip section can go, the last one in file wins is already the expected mango behavior. Nothing bad just my opinion

DreamMaoMao:
This seems to affect all layouts, not just the few currently given ones, and it will make layout calculations very complicated

src/layout/horizontal.h
```diff
@@ -301,61 +327,64 @@ void center_tile(Monitor *m) {
...
+				 ? (m->w.width - cur_gappol - cur_gappor - cur_gappih * ie) *
+					   mfact
+				 : 0;
 
 		if (n - nmasters > 1) {
 			// 多个堆叠窗口：主区域居中，左右两侧各有一个堆叠区域
-			tw = (m->w.width - mw) / 2 - cur_gappoh - cur_gappih * ie;
-			mx = cur_gappoh + tw + cur_gappih * ie;
+			tw = (m->w.width - mw) / 2 - cur_gappol - cur_gappih * ie;
+			mx = cur_gappol + tw + cur_gappih * ie;
```
ernestoCruz05, line 361: hmm wouldnt this still make the right outer gap match gappol? tw is calculated using only cur_gappol and is then reused as the width of the right stack. with gappol=10 and gappor=30, the right stack still ends ~10px from the right edge, same thing with the branch bellow, the if (config.center_when_single_stack)

ernestoCruz05:
> This seems to affect all layouts, not just the few currently given ones, and it will make layout calculations very complicated

i think it would be a neat feature, maybe an issue to see how popular the demand is?

Wateir:
> This seems to affect all layouts

Yeah, it’s the goal to have all layouts affect.

> not just the few currently given ones

do you mean some layouts are one the code but users can’t use them so need to be untouch ?

> and it will make layout calculations very complicated

i don’t see where, since the only change on calculation are some multiplication become substraction or it’s another thing i miss

PR #1284, opened 2026-08-14
feat: add move_client

Sora3e8:
Inspired by previous pull request https://github.com/mangowm/mango/pull/1022
Previously requested in Feature request https://github.com/mangowm/mango/issues/942

Changes done in comparison to #1022
- Synced with current codebase
- Added handling for cross-monitor move
- Simplified the implementation making it as symmetric for dwindle layout and for other layouts as possible

Spec description:
- When move_client,dir is sent window is moved into that direction
- If window hits and edge of a monitor the window will be moved to next monitor if a monitor exists in that direction,
otherwise nothing will happen.

Note: 
- I also attempted to make fully infinitely cycling through monitors if no other monitor is in that direction,
so it would go to x=0 or y=0 again, but the conditions proved troublesome and probably would be good to have a new config for that as well... so I stuck with blocking if no other monitor is in the direction.
- This is nowhere near perfect or complete as some layouts do not work as well with this,
this seems to be primarily caused by way which the functions retrieving client in certain direction work.


Wateir:
https://github.com/mangowm/mango/blob/main/.github/CONTRIBUTING.md

Squash all commit into one, and idk if need run `./format.sh` if you have forgot

Sora3e8:
@Wateir 
I've squashed the commits. 

Regarding the format, I was running the format script through out the process, it's possible I might have forgot to trigger it.
However I've checked the script and it seems to be triggering clang format, I am using neovim with clang as well and I see in the repo there's .clang-format file, so it should be fine.

I tried to run it just in case and I do not see any new diff.

cebem1nt:
Good stuff!

src/dispatch/bind.c
```diff
@@ -173,6 +173,40 @@ void exchange_client(const Arg *arg) {
...
+	}
+
+	Client *tc = direction_select(arg);
+	tc = get_focused_stack_client(tc, arg->tc);
+
+	uint32_t layout_id =
+		(!tc ? monitor_from_direction(arg->i)
+				   ->pertag->ltidxs[c->mon->pertag->curtag]
+				   ->id
+			 : tc->mon->pertag->ltidxs[c->mon->pertag->curtag]->id);
```
DreamMaoMao: This logic is very strange. There's an extra one !tc, why not just tc directly?

in addition，tc->mon->pertag->ltidxs[c->mon->pertag->curtag] should use get_mon_curtag(m) to instead of `c->mon->pertag->curtag`

Issue #852, opened 2026-04-15
Add animation slide from the side


adho08:
In layerrules, there is the option to set the animation to "slide":
`layerrule=animation_type_open:slide,layer_name:notifications`

The animation let's the notification come from the top but I want it to come from the side (namely from the right side). It'd be nice to specify the direction of the "slide" animation.

Wateir:
https://github.com/mangowm/mango/blob/b9c6a2c1964cb59536c5d3e02d2ff981a7eb36d2/src/animation/client.h#L101-L150
This suggest for me it's already implement on the code, just need to be possible to configurable, i have not take enought time to see at what point is realizable

PR #1292, opened 2026-08-18
feat: add animation_direction layer rule

cebem1nt:
Adds animation_direction layer rule, takes in a direction (up/left/right/down). Works with slide animation type.

```
layerrule = animation_direction:right,  layer_name:swaync-control-center
```

Solves #852


cebem1nt:
Misc files were changed after running `format.sh`

cebem1nt:
Should I implement the same window rule? If there is such rule for layers, why not add for windows as well?

PR #1298, opened 2026-08-19
chore: Allow comment after value on the config

### LINKED ISSUE #1295: bug: colors do not handle comments correctly
## Checklist

## Info

If you set any color as:

```conf
bordercolor = 0x0A0A0DDB # And put any comment here
```

It will produce an error, because it takes in comment as part of the value:

```
[ERROR]: Invalid bordercolor format: 0x0A0A0DDB # And put any comment here
╰─[Index] ~/.config/mango/colors/colors.conf:6
   ╰─bordercolor = 0x0A0A0DDB # And put any comment here
```

Wateir:
fix https://github.com/mangowm/mango/issues/1295

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -575,6 +575,25 @@ void trim_whitespace(char *str) {
 	}
 }
 
+// Helper function to strip a trailing comment from a config value.
+void remove_comment(char *str) {
+	if (str == NULL || *str == '\0')
+		return;
+	char *comment = NULL;
+	if (str[0] == '#') {
+		comment = str;
+	} else {
+		for (char *p = str + 1; *p != '\0'; p++) {
+			if (*p == '#' && isspace((unsigned char)p[-1])) {
+				comment = p;
+				break;
+			}
+		}
+	}
+	if (comment != NULL)
+		*comment = '\0';
+}
+
 int32_t parse_double_array(const char *input, double *output,
 						   int32_t max_count) {
 	char *dup = strdup(input);
@@ -3468,6 +3487,7 @@ bool parse_config_line(Config *config, const char *line, int line_number) {
 	}
 
 	// Then trim each part separately
+	remove_comment(value);
 	trim_whitespace(key);
 	trim_whitespace(value);
 
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -575,6 +575,31 @@ void trim_whitespace(char *str) {
 	}
 }
 
+// remove comment, support double quote inside "#xxx" or '#xxx' not be treated
+// as comment
+void remove_comment(char *str) {
+	if (str == NULL || *str == '\0')
+		return;
+
+	char quote_char = '\0';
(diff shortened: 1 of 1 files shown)


cebem1nt (approved):

cebem1nt:
Yeah, fixes it

ernestoCruz05:
What about regex expressions?

cebem1nt:
> What about regex expressions?

Mango uses perl like regexes, in perl coments also begin with `#` so it shouldn't break anything.

if you need to match the `#` exactly, prefix it with `\` :`\#`

DreamMaoMao:
If the spawn parameter already contains # by itself? For example, a hexadecimal string

cebem1nt:
> If the spawn parameter already contains # by itself? For example, a hexadecimal string

I think in such case `spawn_shell` can be used, and a param can be pased like `"#fff"`.

This implementation will look for a space before, but will find a `"`

Wateir:
> If the spawn parameter already contains # by itself? For example, a hexadecimal string

This is a good concern

`bordercolor=#fff` don't gonna trigger it but `bordercolor = #fff` need to be something that work too

I can add something in parse_color so if the first none space char is a `#` keep it

cebem1nt:
> > If the spawn parameter already contains # by itself? For example, a hexadecimal string
> 
> This is a good concern
> 
> `bordercolor=#fff` don't gonna trigger it but `bordercolor = #fff` need to be something that work too
> 
> I can add something in parse_color so if the first none space char is a `#` keep it

but colors are defined in 0xRRGGBBAA hex format, so classic # hex is not valid

Btw, that would be nice to pass raw value that begins with hash if it is prefixed with `\`

Wateir:
So this is a none issue ?
or do this worry you for later ?

cebem1nt:
> So this is a none issue ? or do this worry you for later ?

Yeah, it is not an issue for now

Wateir:
If this is a issue for you to merge, i'm happy to take you're suggestion and make the change to fix this Mao

DreamMaoMao:
```
bind = alt,a,spawn,grim -f -b #00000050 -C #C59D3Eff -w 3 -t ppm
```
What I'm talking about is this situation.

cebem1nt:
> ```
> bind = alt,a,spawn,grim -f -b #00000050 -C #C59D3Eff -w 3 -t ppm
> ```
> 
> What I'm talking about is this situation.

What if wrapping into `""`

```
 bind = alt,a,spawn,grim -f -b "#00000050" -C "#C59D3Eff" -w 3 -t ppm
```
Will it be parsed correctly?

Wateir:
> ```
> bind = alt,a,spawn,grim -f -b #00000050 -C #C59D3Eff -w 3 -t ppm
> ```
> 
> What I'm talking about is this situation.

This is a valid concern, the solution @cebem1nt propose gonna work but hard to find, so i’m gonna make another pr for some doc

Issue #1293, opened 2026-08-18
Expand $HOME in env=


cebem1nt:
Currently there is no way of expanding $HOME or ~/ when setting environment variables. This would be useful because:

Mango does not import environment variables from `.bashrc`, `.zshrc`, so you have to set them manually with `env` however some environment variables require path to the home dir: for example [xdg base directories](https://wiki.archlinux.org/title/)

```conf
env = XDG_STATE_HOME,                $HOME/.local/state
env = XDG_DATA_HOME,                 $HOME/.local/share
env = XDG_CONFIG_HOME,               $HOME/.config
env = XDG_CACHE_HOME,                $HOME/.cache
```

PR #1299, opened 2026-08-20
feat: add ~/ substitution in env

cebem1nt:
#1293, but expands ~/ instead

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -2954,9 +2954,57 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
 		trim_whitespace(env_value);
 
 		ConfigEnv *env = calloc(1, sizeof(ConfigEnv));
-		env->type = strdup(env_type);
-		env->value = strdup(env_value);
 
+		const char *needle = "~/";
+		
+		if (strstr(env_value, needle)) {
+			const char *home = getenv("HOME");
+
+			size_t rlen = 1; // replace only ~
+
+			if (!home) {
+				free(env);
+				mango_error(false, WLR_ERROR,
+							"HOME environment "
+							"variable not set.\n");
+				return false;
+			}
+
+			size_t hlen = strlen(home);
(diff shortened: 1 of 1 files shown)


src/config/parse_config.h
```diff
@@ -2954,9 +2954,52 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
...
+		const char *home = getenv("HOME");
+		const char *needle = "~/";
+
+		size_t rlen = 1; // replace only ~
+
+		if (!home) {
+			mango_error(false, WLR_ERROR,
+						"HOME environment "
+						"variable not set.\n");
+			return false;
```
DreamMaoMao: Returning directly here will cause a memory leak

src/config/parse_config.h
```diff
@@ -2954,9 +2954,52 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
...
+		size_t hlen = strlen(home);
+		size_t len = strlen(env_value) + 1;
+
+		for (char *p = strstr(env_value, needle); p;
+			 p = strstr(p + rlen, needle))
+			len += hlen - rlen;
+
+		env->value = malloc(len);
+		if (!env->value) {
+			free(env->type);
```
DreamMaoMao: A null pointer is released here

src/config/parse_config.h
```diff
@@ -2954,9 +2954,52 @@ bool parse_option(Config *config, char *key, char *value, int line_number) {
 		trim_whitespace(env_value);
 
 		ConfigEnv *env = calloc(1, sizeof(ConfigEnv));
-		env->type = strdup(env_type);
-		env->value = strdup(env_value);
 
+		const char *home = getenv("HOME");
+		const char *needle = "~/";
```
DreamMaoMao: It is necessary to determine whether ~ is included before reading the environment variable; otherwise, all parsing will read the environment variable.

cebem1nt:
my bad, fixed now

PR #1302, opened 2026-08-21
feat: thumbnail switcher across all tags and monitors

stepbrobd:
switcher will show live preview for all windows across all workspace tags and monitors

Screenshot from 2026-08-21 14-13-00

### DIFF
--- a/src/input/keyboard.h
+++ b/src/input/keyboard.h
@@ -508,6 +508,18 @@ void keypress(struct wl_listener *listener, void *data) {
 		toggleoverview(&(Arg){.i = 1});
 	}
 
+	if (switcher_is_active()) {
+		if (locked) {
+			switcher_close();
+		} else if (!group->virtual_keyboard &&
+				   event->state == WL_KEYBOARD_KEY_STATE_RELEASED &&
+				   ISMODEKEYCODE(keycode)) {
+			switcher_commit();
+			group->nsyms = 0;
+			wl_event_source_timer_update(group->key_repeat_source, 0);
+		}
+	}
+
 	if (config.cursor_hide_on_keypress && !cursor_hidden &&
 		event->state == WL_KEYBOARD_KEY_STATE_PRESSED) {
 		hidecursor(NULL);
--- a/src/input/pointer.h
+++ b/src/input/pointer.h
@@ -441,6 +441,19 @@ bool handle_buttonpress(struct wlr_pointer_button_event *event) {
 		if (locked)
 			break;
 
+		if (switcher_is_active() &&
+			(event->button == BTN_LEFT || event->button == BTN_RIGHT)) {
+			Client *switcher_c = switcher_client_at(cursor->x, cursor->y);
+			if (!switcher_c)
+				switcher_close();
+			else if (event->button == BTN_LEFT)
+				switcher_commit_client(switcher_c);
+			else
+				pending_kill_client(switcher_c);
+			wlr_seat_pointer_notify_clear_focus(seat);
+			return true;
+		}
+
 		xytonode(cursor->x, cursor->y, &surface, NULL, NULL, &gb, NULL, NULL);
 		if (toplevel_from_wlr_surface(surface, &c, &l) >= 0) {
 			if (c && c->scene && c->scene->node.enabled &&
--- a/docs/bindings/keys.md
+++ b/docs/bindings/keys.md
@@ -118,6 +118,7 @@ bindr=Super,Super_L,spawn,rofi -show run
 | `focus_window_or_workspace` | `left/right/up/down` | Focus window in direction; otherwise jump to the nearest adjacent tag that has clients, falling back to the next/previous tag if none. |
 | `focusstack` | `next/prev` | Cycle focus within the stack. |
 | `focuslast` | - | Focus the previously active window. |
+| `switcher` | `next/prev` | Open or cycle the all-tag, all-monitor thumbnail switcher. Releasing any modifier key selects. |
 | `exchange_client` | `left/right/up/down` | Swap window with neighbor in direction. |
 | `exchange_stack_client` | `next/prev` | Exchange window position in stack. |
 | `zoom` | - | Swap focused window with Master. |
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -702,6 +702,7 @@ static void set_binding_keymode(Config *config, char mode[28],
 
 int32_t parse_circle_direction(const char *str) {
 	// 将输入字符串转换为小写
+
 	char lowerStr[10];
 	int32_t i = 0;
 	while (str[i] && i < 9) {
@@ -1298,6 +1299,9 @@ FuncType parse_func_name(char *func_name, Arg *arg, char *arg_value,
 		func = centerwin;
 	} else if (strcmp(func_name, "focuslast") == 0) {
 		func = focuslast;
+	} else if (strcmp(func_name, "switcher") == 0) {
+		func = switcher;
+		(*arg).i = parse_circle_direction(arg_value);
 	} else if (strcmp(func_name, "toggle_trackpad_enable") == 0) {
 		func = toggle_trackpad_enable;
 	} else if (strcmp(func_name, "setoption") == 0) {
--- a/src/dispatch/bind_declare.h
+++ b/src/dispatch/bind_declare.h
@@ -6,6 +6,7 @@ void focus_window_or_workspace(const Arg *arg);
 void groupjoin(const Arg *arg);
 void groupleave(const Arg *arg);
 void toggleoverview(const Arg *arg);
+void switcher(const Arg *arg);
 void togglehdr(const Arg *arg);
 void togglejump(const Arg *arg);
 void set_proportion(const Arg *arg);
--- a/src/manage/client.h
+++ b/src/manage/client.h
@@ -2027,6 +2027,7 @@ void unmapnotify(struct wl_listener *listener, void *data) {
 	Monitor *m = NULL;
 	Client *nextfocus = NULL;
 	c->iskilling = 1;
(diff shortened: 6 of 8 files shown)


Wateir:
Why use that over overview ?

stepbrobd:
overview only work per workspace tag (doesnt work across different tags and monitors) 

i want to have a switcher for all windows across monitors and tags

the screenshot is showing newly added (WIP) switcher on 2 monitors across 5 workspaces

DreamMaoMao:
It seems that the content of the thumbnails cannot be seen clearly. They have been scaled too small.

If you can make them fully occupy the entire screen and then rename it workspace overview, I would consider merging

Wateir:
Can you share a new pictures of what it’s look like after the feedback of mao ?

stepbrobd:
> It seems that the content of the thumbnails cannot be seen clearly. They have been scaled too small.

IMO i want to have this feature like windows/macos window switcher. in windows/macos, the switcher works across monitors and tags and is not intended to be used for viewing actual content inside the thumbnails (more like only showing the rough shape of each window) so i think scaling them to be very small exactly what i wanted them to be

> If you can make them fully occupy the entire screen and then rename it workspace overview, I would consider merging

i assume you don't want a tab switcher like how niri's SUPER+Tab works but like a overview (as in the current mango overview implementation) but grab windows from other tags and monitors in full screen with animation? implementation wise this will be a lot easier/smaller but from UX standpoint it might be a bit too overwhelming

> Can you share a new pictures of what it’s look like after the feedback of mao ?

i'll be a bit busy next week, so i want to get Mao's opinion before making another big edit (the past few pushes are just rebasing)

P.S. i've already bought a some crypto for sponsoring the project but they are not settled for transfer yet (coinbase bruh), maybe it'd be easier for potential sponsors if there are more accessible ways of accepting donations (maybe GitHub sponsors or Open Collective?)

DreamMaoMao:
> i assume you don't want a tab switcher like how niri's SUPER+Tab works but like a overview (as in the current mango overview implementation) but grab windows from other tags and monitors in full screen with animation? implementation wise this will be a lot easier/smaller but from UX standpoint it might be a bit too overwhelming

It's not like that. I just simply think the display is too small. There's still a lot of space available. Why not use them to be a bit larger?

oh, I think I was mistaken. Right now, it's just a preview of the displayed window, not the workspace.

> P.S. i've already bought a some crypto for sponsoring the project but they are not settled for transfer yet (coinbase bruh), maybe it'd be easier for potential sponsors if there are more accessible ways of accepting donations (maybe GitHub sponsors or Open Collective?)

Thank you for your sponsorship. It's kind helpful. However, due to the limitations of my environment, I can only accept donations in this way for now.

stepbrobd:
is this version better?

Screenshot from 2026-08-23 17-08-49

Screenshot from 2026-08-23 17-09-50

Screenshot from 2026-08-23 17-10-30

you might want to take a look at the margin/padding/gaps etc (i just picked some numbers feels right to me) and these are not configurable because i think the current configuration surface is already big enough...

preferred thumbnail height is 1/4 of the focused monitor's logical height. Its reduced as needed to keep the panel within 80% of the monitor width/height. Thumbnail width follows the source window's aspect ratio (clamped between 9:16 and 12:5) and windows outside that aspect ratio are stretched

in the current implementation we takes a zero copy snapshot on switcher opens and retains current buffers for the root surface and existing subsurfaces, then releases them when the panel closes (note that later commits after switcher open do NOT update the thumbnails)

btw i think its possible to add live preview for each thumbnail (i asked my friend on windows 11 and they told me the tab switcher on windows have live preview), maybe extra 200 LoC?

Wateir:
Can you put the magic number outside of the function and on some macros ? You are right it's already a lot of config thing, but for people like me who use our own fork we can easly change them

DreamMaoMao:
At present, it has inserted special logic in too many places. I think a lot of it is unnecessary and can be handled centrally. For instance, why does the release detection need to be bound to specific keys on a specific keyboard? Wouldn't it be better to just monitor the release of any mod key like in ov tab mode?

I believe its logic should only appear in keypress and unmapnotify.

src/input/keyboard.h
```diff
@@ -216,6 +216,7 @@ static void create_standalone_keyboard(InputDevice *input_dev,
 
 void destroy_standalone_keyboard(struct wl_listener *listener, void *data) {
 	KeyboardGroup *group = wl_container_of(listener, group, destroy);
+	switcher_drop_keyboard(group->keyboard);
```
DreamMaoMao: don't do this.

src/input/keyboard.h
```diff
@@ -302,6 +303,7 @@ KeyboardGroup *createkeyboardgroup(void) {
 
 void destroykeyboardgroup(struct wl_listener *listener, void *data) {
 	KeyboardGroup *group = wl_container_of(listener, group, destroy);
+	switcher_drop_keyboard(group->keyboard);
```
DreamMaoMao: don't do this.

src/manage/misc.h
```diff
@@ -349,6 +349,7 @@ void locksession(struct wl_listener *listener, void *data) {
 		return;
 	}
 	lock = session_lock->data = ecalloc(1, sizeof(*lock));
+	switcher_cancel();
```
DreamMaoMao: dont do tihs.

src/manage/monitor.h
```diff
@@ -720,6 +720,8 @@ void closemon(Monitor *m) {
 		toggleoverview(&(Arg){.i = 1});
 	}
 
+	switcher_cancel();
```
DreamMaoMao: dont do this.

src/dispatch/bind_define.h
```diff
@@ -1920,6 +1920,8 @@ void toggleoverview(const Arg *arg) {
 	if (!selmon || grabc)
 		return;
 
+	switcher_cancel();
```
DreamMaoMao: dont do this.

DreamMaoMao (commented):
To disable its display, simply monitor any mod key being released. All other situations are extreme usage scenarios, and it can also be made to disappear by triggering it again. There is no need to insert special logic everywhere in the code to cover up

stepbrobd:
i've removed switcher specific hooks (`keypress` and `unmapnotify` only as you mentioned)

src/input/keyboard.h
```diff
@@ -508,6 +508,28 @@ void keypress(struct wl_listener *listener, void *data) {
...
+			wl_event_source_timer_update(group->key_repeat_source, 0);
+		} else if (event->state == WL_KEYBOARD_KEY_STATE_PRESSED) {
+			for (i = 0; i < nsyms; i++) {
+				if (syms[i] != XKB_KEY_Escape)
+					continue;
+				switcher_close();
+				// stop a held opener from reopening the switcher
+				group->nsyms = 0;
+				wl_event_source_timer_update(group->key_repeat_source, 0);
+				return;
```
DreamMaoMao: What is the significance of adding this esc key? Can't it be turned off simply by triggering any mod keys?

src/input/pointer.h
```diff
@@ -441,6 +441,19 @@ bool handle_buttonpress(struct wlr_pointer_button_event *event) {
 		if (locked)
 			break;
 
+		if (switcher_is_active() &&
+			(event->button == BTN_LEFT || event->button == BTN_RIGHT)) {
+			Client *switcher_c = switcher_client_at(cursor->x, cursor->y);
+			if (!switcher_c)
+				switcher_close();
+			else if (event->button == BTN_LEFT)
+				switcher_commit_client(switcher_c);
```
DreamMaoMao, line 450: Is it really necessary to select this kind of switcher with a pointer?

stepbrobd: i was trying to replicate overview's pointer event handling

src/input/keyboard.h
stepbrobd: i think this might be a UX problem

currently in both overview and switcher, if you esc while switcher/overview is open, the event gets sent to the focused client

stepbrobd: if you think this is a non issue i can drop the handler

stepbrobd:
just donated $66.00 (apparently BNB has gone up a little since i bought it)

https://bscscan.com/tx/0x921414e6c5ec120d6e70641d8cf8552f8367c41b92b88935c2a1c5aa6e2a7ffb

thanks a lot for making mango!

if you need a Nix CI and a binary cache please lmk, I'm happy to host this for the community :) https://github.com/stepbrobd/atelier

src/input/keyboard.h
DreamMaoMao: > currently in both overview and switcher, if you esc while switcher/overview is open, the event gets sent to the focused client

I think it's best to delete this. It's an overly special usage scenario.

src/input/pointer.h
DreamMaoMao: all right

DreamMaoMao:
> just donated $66.00 (apparently BNB has gone up a little since i bought it)
> 
> https://bscscan.com/tx/0x921414e6c5ec120d6e70641d8cf8552f8367c41b92b88935c2a1c5aa6e2a7ffb
> 

thanks you.

>
> if you need a Nix CI and a binary cache please lmk, I'm happy to host this for the community :) https://github.com/stepbrobd/atelier

This is not needed for the time being.

src/input/keyboard.h
stepbrobd: dropped

stepbrobd:
esc handler is dropped, are there any other things i need to fix/change?

DreamMaoMao:
Its logic when killing a client is a bit strange. Each time it kills, it has to traverse all the thumbnails for comparison, but then clean them all up in the end

PR #1303, opened 2026-08-21
ipc: fix focusid not working, add new dispatchers

cebem1nt:
Adds more dispatchers to interact with clients by ID

- [x] fix focusid
- [x] add killid
- [x] add togglefloatingid
- [x] add togglefullscreenid
- [x] add tagid

## Why?

I'm trying to port nwg-dock-hyprland to mango, it lacks these

### DIFF
--- a/src/config/parse_config.h
+++ b/src/config/parse_config.h
@@ -719,8 +719,12 @@ int32_t parse_direction(const char *str) {
 
 int32_t parse_force(const char *str) {
 	// 将输入字符串转换为小写
+	if (!str)
+		return UNFORCE;
+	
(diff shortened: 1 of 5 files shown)


DreamMaoMao:
mmsg dispatch togglefloating client,1

why not use those?

cebem1nt:
> mmsg dispatch togglefloating client,1
> 
> why not use those?

- It was not documened, didn't know it exists
- It is confusing, I have focusid to focus by id, but dont have alternatives for kill, tag, etc

but well, fair enough I guess

Edited: Ah, hell, I see now

PR #1306, opened 2026-08-22
fix: make floating windows appear behind layers

cebem1nt:
This PR fixes floating windows conflicting with layer surfaces and full screened windows on the appearance layer. Forces floating windows to be behind full screened windows and layers


cebem1nt:
implemented in b07190dad1743119739695786e606013ca5de1b9, thx!

PR #1331, opened 2026-08-29
(nix) added clang-tools to nix flake devshell

SDG-Den:
This change allows format.sh to be run from the devshell (nix develop) without needing clang-tools installed globally on the system.


SDG-Den:
resolved

Issue #450, opened 2025-11-24
[Feature request] animations for showing/hiding scratchpads


cerebroso:
It would be great to have animations for showing/hiding a scratchpad, so we could have yakuake style terminals. If we could have them sliding from/to a direction as well, it would be perfect! (like sliding from the top)

ellaweh:
Would be nice indeed!!

og900aero:
Uhh, I'd like to set something like that up too, but there's only the option to open and close the program

cebem1nt:
Yeah, this would be very nice!

DreamMaoMao:
done

PR #1332, opened 2026-08-29
feat: add animations for minimizing, unminimizing and scratchpad

eestsoftware:
#450 
Added animations that play when a window is minimized, restored from being minimized, or the scratchpad is toggled. Uses the same animations as open and close, (i.e. slide, zoom, fade, none). 

How it is defined in config.conf:
``` 
animation_type_unminimize=zoom (default)
animation_type_minimize=fade (default)

animation_duration_unminimize=350 (default)
animation_duration_minimize=300 (default)

animation_curve_unminimize=0.46,1.0,0.29,0.99 (default)
animation_curve_minimize=0.46,1.0,0.29,0.99 (default)
```

Animation types can also be applied in window rules.

### DIFF
--- a/assets/config.conf
+++ b/assets/config.conf
@@ -79,7 +79,6 @@ dwindle_preserve_split=0
 # Overview Setting
 hotarea_size=10
 enable_hotarea=0
-ov_tab_mode=1
 overviewgappi=5
 overviewgappo=30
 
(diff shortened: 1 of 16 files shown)


DreamMaoMao:
57be367ae7257f346cb5e8aae0055b7cfa4a96f3

done in this.

PR #1345, opened 2026-09-01
fix: restore seat keyboard after active keyboard destruction

### LINKED ISSUE #1344: Destroying a transient virtual keyboard leaves wlr_seat without an active keyboard
## Checklist

## Info

mango version (`mango -v`): `mango 0.16.2(53d729c0)`
wlroots version: `0.20.2`
Wayland version: `1.26.0`

## Crash track

Mango itself does not crash. Affected clients may crash after receiving keyboard enter/modifier events without first receiving a keymap.

For example, Vesktop using Electron 43 crashes in this path:

```text
xkb_state_update_mask
ui::XkbKeyboardLayoutEngine::UpdateModifiers
ui::WaylandKeyboard::OnModifiers
```

## Description

Destroying a transient virtual keyboard can leave `wlr_seat` without an active keyboard while Mango continues to advertise `WL_SEAT_CAPABILITY_KEYBOARD`.

A client which binds `wl_keyboard` afterwards does not receive `wl_keyboard.keymap`. Once the client is focused, it still receives `wl_keyboard.enter` and `wl_keyboard.modifiers`. This caused Vesktop/Electron 43 to crash and QQ/Electron 40 to display a blank window in my real session.

A physical keyboard event can temporarily hide the problem by calling `wlr_seat_set_keyboard()` again. Reloading Mango also restores the keyboard.

## Steps to reproduce

A deterministic headless reproducer is available here:

https://gist.github.com/FlyongX/0179f8eaa5ddc6b35333abe34508392b

It starts Mango with the headless backend, creates and destroys a short-lived virtual keyboard via `wtype`, and then connects a fresh client which checks whether `wl_keyboard.keymap` is received:

```bash
./test-virtual-keyboard-keymap.sh /path/to/mango
```

Result on unmodified `main` at `53d729c0`:

```text
keymap=0
```

The test exits with status 1. I reproduced this on every run.

Expected result:

```text
keymap=1
```

## Protocol sequence

Before the virtual keyboard is destroyed, a newly connected client receives:

```text
wl_seat.get_keyboard
wl_keyboard.keymap
wl_keyboard.repeat_info
wl_keyboard.enter
wl_keyboard.modifiers
```

After it is destroyed, a newly connected client receives:

```text
wl_seat.get_keyboard
# no wl_keyboard.keymap
wl_keyboard.enter
wl_keyboard.modifiers
```

## Cause

`virtualkeyboard()` creates a separate keyboard group. `createkeyboardgroup()` makes that group the seat keyboard with `wlr_seat_set_keyboard()`:

https://github.com/mangowm/mango/blob/53d729c0476fc50534835dcdaf58ad314a32a34d/src/input/keyboard.h#L232-L300

When the virtual keyboard is destroyed, `destroykeyboardgroup()` destroys its `wlr_keyboard_group`. wlroots consequently clears `seat->keyboard_state.keyboard`, but Mango does not restore the persistent physical-keyboard group (`kb_group`):

https://github.com/mangowm/mango/blob/53d729c0476fc50534835dcdaf58ad314a32a34d/src/input/keyboard.h#L303-L313

Restoring `kb_group->keyboard` after `wlr_keyboard_group_destroy()` makes the reproducer return `keymap=1`; I will submit the tested fix as a linked pull request.

FlyongX:
## Summary

- Restore the persistent keyboard group (`kb_group`) after an active virtual-keyboard group is destroyed.
- Apply the same fallback when an active standalone keyboard is removed.
- Defer restoration to an idle callback; never overwrite another keyboard that became active during destruction.
- Identify virtual keyboard groups via `group->virtual_keyboard`, as requested in review.
- Keep the lifecycle rationale as comments rather than removing it.

Fixes #1344.

## Why

On current `main`, `createkeyboardgroup()` no longer replaces an existing seat keyboard, which fixed the virtual keyboard that is created and destroyed without sending input.

A virtual keyboard that sends a key or modifier still becomes the active seat keyboard through `keypress()` / `keypressmod()`. When that short-lived group is destroyed, `destroykeyboardgroup()` clears the seat keyboard but does not restore `kb_group->keyboard`. Mango keeps advertising `WL_SEAT_CAPABILITY_KEYBOARD`, so a client binding `wl_keyboard` afterwards receives no keymap and can later receive enter/modifier events without an XKB state (Electron crash / blank window).

The same lifecycle boundary exists for standalone keyboards created by keyboard-specific `devicerule` rules: they are not members of `kb_group`, may own the seat directly, and `destroy_standalone_keyboard()` also clears the seat without restoring the persistent group.

## What this change does

In both destroy handlers, record whether the keyboard being destroyed was the active seat keyboard:

- If it was active, clear the seat as before, then schedule an idle callback.
- The idle callback runs after all keyboard destroy listeners. It restores `kb_group->keyboard` only if the seat is still empty, so it never overwrites another keyboard activated during destruction.
- For the default `kb_group` itself (shutdown), no callback is scheduled, so the shutdown path is unchanged.

This is an immediate fallback policy for the confirmed bug. It does not attempt to choose between multiple still-live virtual keyboards; that can be investigated separately once the seat can no longer be left without a keymap.

## Testing

Tested against `main` at `a092a34aff906c6c53593b1ef8876fcd236fc1a1` with wlroots 0.20.2.

- `./format.sh`
- `git diff --check`
- regular Meson/Ninja build
- ASAN/UBSAN Meson/Ninja build
- headless `wl_keyboard.keymap` protocol reproducer (empty, key, and modifier virtual keyboards)

Current `main` before this change:

| scenario | result |
| --- | --- |
| empty virtual keyboard | `keymap=1` |
| virtual keyboard sends a key | `keymap=0` |

With this change:

| scenario | result |
| --- | --- |
| empty virtual keyboard | `keymap=1` (5/5) |
| virtual keyboard sends a key | `keymap=1` (5/5) |
| virtual keyboard sends a modifier | `keymap=1` |
| ASAN/UBSAN, empty and key | `keymap=1` |

The standalone keyboard path uses the same helper and is covered by the build and the fallback logic above; it is not exercised by the headless reproducer and has not been reproduced with a physical hot-unplug device.

### DIFF
--- a/src/input/keyboard.h
+++ b/src/input/keyboard.h
@@ -221,13 +221,17 @@ static void create_standalone_keyboard(InputDevice *input_dev,
 	input_dev->device_data = group;
 }
 
+static void restore_default_seat_keyboard(void *data) {
+	if (seat && kb_group && !wlr_seat_get_keyboard(seat))
+		wlr_seat_set_keyboard(seat, kb_group->keyboard);
+}
+
 void destroy_standalone_keyboard(struct wl_listener *listener, void *data) {
 	KeyboardGroup *group = wl_container_of(listener, group, destroy);
 	if (group->keyboard == last_active_keyboard)
 		last_active_keyboard = NULL;
-	// devicerule 的键盘不在 kb_group 里，拔掉时要是正占着 seat 就悬空，
-	// 反正下一次按键会重新设置
-	if (wlr_seat_get_keyboard(seat) == group->keyboard)
+	bool restore_keyboard = wlr_seat_get_keyboard(seat) == group->keyboard;
+	if (restore_keyboard)
 		wlr_seat_set_keyboard(seat, NULL);
 	wl_list_remove(&group->key.link);
 	wl_list_remove(&group->modifiers.link);
@@ -238,6 +242,8 @@ void destroy_standalone_keyboard(struct wl_listener *listener, void *data) {
 	}
 	wl_list_remove(&group->link);
 	free(group);
+	if (restore_keyboard)
+		wl_event_loop_add_idle(event_loop, restore_default_seat_keyboard, NULL);
 }
 
 KeyboardGroup *createkeyboardgroup(void) {
@@ -313,8 +319,8 @@ void destroykeyboardgroup(struct wl_listener *listener, void *data) {
 	KeyboardGroup *group = wl_container_of(listener, group, destroy);
 	if (group->keyboard == last_active_keyboard)
 		last_active_keyboard = NULL;
-	// 销毁的是整个组，组的键盘也跟着没了。它要是正占着 seat 就悬空，
-	// 后面按键会重新设置
+	bool restore_keyboard = group->virtual_keyboard &&
+							wlr_seat_get_keyboard(seat) == group->keyboard;
 	if (wlr_seat_get_keyboard(seat) == group->keyboard)
 		wlr_seat_set_keyboard(seat, NULL);
 	wl_event_source_remove(group->key_repeat_source);
@@ -323,6 +329,8 @@ void destroykeyboardgroup(struct wl_listener *listener, void *data) {
 	wl_list_remove(&group->destroy.link);
 	wlr_keyboard_group_destroy(group->wlr_group);
 	free(group);
+	if (restore_keyboard)
+		wl_event_loop_add_idle(event_loop, restore_default_seat_keyboard, NULL);
 }
 
 int32_t keyrepeat(void *data) {
--- a/src/input/keyboard.h
+++ b/src/input/keyboard.h
@@ -27,19 +27,26 @@ void createkeyboard(struct wlr_keyboard *keyboard) {
 
 	ConfigDeviceRule *rule = find_device_rule(&keyboard->base);
 	if (rule && device_rule_has_keyboard_settings(rule) && input_dev) {
-		/* 命中带键盘参数的 devicerule：独立创建，使用独立 keymap */
+		// 命中带键盘参数的 devicerule：独立创建，使用独立 keymap
 		input_dev->standalone = true;
 		create_standalone_keyboard(input_dev, keyboard, rule);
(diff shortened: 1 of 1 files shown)


src/input/keyboard.h
```diff
@@ -309,6 +309,11 @@ void destroykeyboardgroup(struct wl_listener *listener, void *data) {
 	wl_list_remove(&group->modifiers.link);
 	wl_list_remove(&group->destroy.link);
 	wlr_keyboard_group_destroy(group->wlr_group);
+	/* Destroying the active virtual-keyboard group clears the seat keyboard.
+	 * Restore the persistent group so new wl_keyboard resources get a keymap.
+	 */
+	if (group != kb_group && !wlr_seat_get_keyboard(seat))
```
DreamMaoMao: group->virtual_keyboard

  should use this to judge whethe the group is virtual keyboard.

FlyongX: Updated. The virtual-group restoration path now uses `group->virtual_keyboard`, and it only schedules the fallback when that virtual group was the active seat keyboard. The idle callback also checks that the seat is still empty before restoring `kb_group->keyboard`, so it will not overwrite another keyboard activated during destruction.

I also retained and expanded the lifecycle comments around both virtual-group and standalone-keyboard destruction instead of removing the original rationale.

FlyongX:
Closing this PR because the upstream fix in
`f875e194eddb9f19f243c7977ad4b00f874612a9` is now on `main`.

I tested `main` at `98818a61573f5ece1ea0cb5f98f179ef2061e6eb`. The
headless empty, key-producing, modifier, and multiple-virtual-keyboard
regression paths all return `keymap=1`. After installing it and restarting
the Mango session, I also have not been able to reproduce the previous
Electron crash or blank-window behavior with Fcitx5 enabled.

The separate fallback implementation in this PR is therefore no longer
needed. Thanks for reviewing it and for implementing the upstream fix.

PR #1349, opened 2026-09-02
Feat: Added groupmerge dispatcher and option for group to auto-capture new windows.

SDG-Den:
groupmerge works the other way around from groupjoin:

groupjoin,left will make the focused window join the group to the left.

groupmerge,left will make the window to the left of the focused window join the focused window in a group

I personally use a script that does focusdir,<dir> + groupjoin,<inverse of dir> to do this, since it's really hard to actually group multiple things into a group quickly with groupjoin.

Feat: Added group_capture_spawn setting

defaults to 0, if enabled, new windows will auto-join the focused group when spawned. this matches the behaviour i experienced on cosmic's groups, and felt very intuitive, so i wanted to add it.


SDG-Den:
Feat: added new group dispatchers:
groupinit - initializes the target client as a group with only itself as the client (for use with group_spawn_capture)

groupall - groups all clients

groupdisband - ungroups an entire group at once.

PR #1350, opened 2026-09-03
feat: monocle layout client count, like in dwm

savar95x:
### Summary
**Feature**: Instead of a default `M` for monocle layout, it now gives `[num]`, num being the number of clients open in that tag. This is similar to how dwm displays it's monocle layout, and is very handy if used with a bar like waybar.


savar95x:
Any plans for this?

PR #1351, opened 2026-09-03
fix: delay on launching apps (eg foot) via keybind on non-systemd distros (eg void)

savar95x:
### Summary
**Bugfix**: mangowc on non-systemd distros like void had noticeable delay in launching apps via a keybind. This PR fixes that bug, and has been tested rigorously in my daily use.

Particularly, the `spawn` and `spawn_shell` functions have been updated.

#### Backstory
I had compiled mangowc on void, and there was a noticeable delay in foot launch speed using the keybind, especially compared to running it directly via an existing foot instance. Also, I didn't remember facing such an issue on arch.

### DIFF
--- a/src/dispatch/bind_define.h
+++ b/src/dispatch/bind_define.h
@@ -1117,6 +1117,11 @@ void spawn_shell(const Arg *arg) {
 	const char *activation_token = xdg_activation_v1_export_token();
 
 	if (fork() == 0) {
+		// fix: Explicitly unblock all signals inherited from wlroots
+		sigset_t set;
+		sigemptyset(&set);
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
Could you test the latest commit? It should solve this problem

savar95x:
Works like a charm! Closing this PR.

savar95x:
Hey @DreamMaoMao, the launch delay (for keybind launch of foot on non-systemd distros like void) is back there in the latest build. It feels less duration than the original delay, but this one is still noticeable, probably around 500ms.

Could you fix it again?

DreamMaoMao:
done

PR #1353, opened 2026-09-03
Feat: Added various group utilities

SDG-Den:
Added new option group_capture_spawn.

when enabled, spawned windows when a group is focused will automatically be captured to the group and their open animation will be suppressed. 

also added the following dispatchers:

groupinit - initializes a single window as a group, this makes it work with group_capture_spawn. 
groupmerge,DIR - pull window in direction DIR into the focused window/group
groupall - group all visible windows on the monitor.
groupdisband - disbands a group, returning all clients to being individually tiled.


SDG-Den:
refactored into this pr: https://github.com/mangowm/mango/pull/1376

PR #1354, opened 2026-09-03
Feat: Add Layout Binds

SDG-Den:
This feature i came up with based on some conversations in the discord.

new config option:

layoutbind=layout,event,dispatcher,args

events are: switchto/switchfrom (explicit, only trigger on dispatch events and compositor initialization) and entry/exit (implicit, will also trigger on tag switch and keyboard focus switch)

re-uses the same dispatcher logic from keybinds, and supports all the same dispatchers. 

this allows for layout-specific daemons as well as communications over IPC back to the user's shell when the layout changes. without taking into account external applications, this also allows things like gaps to be toggled for specific modes and allows for layout-specific keymodes.

### DIFF
--- a/docs/bindings/keys.md
+++ b/docs/bindings/keys.md
@@ -89,6 +89,8 @@ bind=none,Super_L,spawn,rofi -show run
 bindr=Super,Super_L,spawn,rofi -show run
 ```
 
+
+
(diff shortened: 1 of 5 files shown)


SDG-Den:
closing pending possible rework.

PR #1356, opened 2026-09-03
feat: implement special tag overlay workspace (scratchpad)

Ssnibles:
## Summary

Adds a special tag overlay workspace (scratchpad) with full tiling support across layouts (Scroller, Master/Stack, Dwindle, Monocle, Center Tile, etc.) Multiple windows assigned to the special tag tile cohesively over the active workspace beneath an interactive dim barrier.

---

## Key Changes

* **Independent Pertag Slot (`0`):** Sized `Pertag` to `tag_num_MAX + 1` and reserved index `0` for the special workspace. Layout, `mfact`, `nmaster`, scroller state, and dwindle trees operate independently from standard tags.
* **Dedicated Scene Layers:** Introduced `LyrSpecialDim`, `LyrSpecialTile`, `LyrSpecialTop`, and `LyrSpecialFullscreen`. Special windows and dim layers render above normal fullscreen windows (`LyrFullscreen`) and below global overlays.
* **Input Barrier:** `LyrSpecialDim` blocks mouse hover and click events in `xytonode()`, preventing clicks through to background windows.
* **Multi-Monitor Handling:** Enforces a single active special overlay across displays via `ensure_single_special_monitor`, migrating special windows to the active monitor (`selmon`).
* **Auto-Dismiss Lifecycle:** Overlay automatically closes when the last special window is dismissed, untagged, or when switching workspaces with `view`.
* **Animations:** Windows slide in/out from the top edge while underlying windows remain static under the dim layer.
* **IPC & Overview Compatibility:** Hides special windows during overview mode and exposes `"is_special": true` in the IPC client JSON payload.

---

### Keybindings

```ini
# Toggle overlay visibility
bind = SUPER, s, toggle_special_tag

# Move focused window to/from special workspace
bind = SUPER+SHIFT, s, tag_special_tag

# Silently send active window to special workspace without shifting view
bind = SUPER+CTRL, s, tag_special_silent

```

### Visuals & Gaps

```ini
# Dimming opacity behind overlay (0.0 to 1.0, default: 0.5)
special_dim = 0.5

# Inner and outer gaps
special_gappih = 10
special_gappiv = 10
special_gappoh = 20
special_gappov = 20

```

### Window & Tag Rules

```ini
# Assign applications directly on launch
windowrule = special_tag:1, appid:spotify
windowrule = special_tag:1, appid:discord

# Default layout for special workspace (id: 0)
tagrule = id:0, layout_name:dwindle

```

---

## Testing & Validation

**Layouts:** Verified proper tiling across Master/Stack, Scroller, Dwindle, and Center Tile.
**Input Handling:** Confirmed `LyrSpecialDim` blocks pass-through hover and click interactions.
**Z-Order:** Verified overlay surfaces cleanly over standard fullscreen windows; fullscreen special windows promote to `LyrSpecialFullscreen`.
**Multi-Monitor:** Confirmed migration to `selmon` and auto-close on previous monitors.
**Lifecycle:** Confirmed overlay auto-dismisses when the last window is closed or moved.
**Rules & IPC:** Verified `special_tag:1` rule matches at launch, overview correctly hides windows, and IPC returns valid `is_special` / `is_special_tag` keys.

---

  Screenshot_2026-09-04_11-46-53
  Screenshot_2026-09-04_11-45-50
  Screenshot_2026-09-04_11-45-11

### DIFF
--- a/src/animation/client.h
+++ b/src/animation/client.h
@@ -60,18 +60,14 @@ struct fx_corner_radii set_client_corner_location(Client *c) {
 }
 
 bool is_horizontal_stack_layout(Monitor *m) {
-	if (m->pertag->curtag &&
-		(m->pertag->ltidxs[m->pertag->curtag]->id == TILE ||
-		 m->pertag->ltidxs[m->pertag->curtag]->id == DECK))
-		return true;
-	return false;
+	uint32_t tag = get_mon_curtag(m);
+	return m->pertag->ltidxs[tag]->id == TILE ||
+		   m->pertag->ltidxs[tag]->id == DECK;
 }
 
 bool is_horizontal_right_stack_layout(Monitor *m) {
-	if (m->pertag->curtag &&
-		m->pertag->ltidxs[m->pertag->curtag]->id == RIGHT_TILE)
-		return true;
-	return false;
+	uint32_t tag = get_mon_curtag(m);
+	return m->pertag->ltidxs[tag]->id == RIGHT_TILE;
 }
 
 int32_t is_special_animation_rule(Client *c) {
@@ -542,7 +538,8 @@ void client_draw_split_border(Client *c, bool hit_no_border,
 	if (c->iskilling || !c->mon || !client_surface(c)->mapped)
 		return;
 
-	const Layout *layout = c->mon->pertag->ltidxs[c->mon->pertag->curtag];
+	uint32_t tag = get_client_tag_idx(c);
+	const Layout *layout = c->mon->pertag->ltidxs[tag];
 
 	if (hit_no_border || !ISTILED(c) || layout->id != DWINDLE ||
 		!config.dwindle_manual_split || c->isfullscreen) {
@@ -553,7 +550,7 @@ void client_draw_split_border(Client *c, bool hit_no_border,
 		return;
 	}
 
-	DwindleNode **root = &c->mon->pertag->dwindle_root[c->mon->pertag->curtag];
+	DwindleNode **root = &c->mon->pertag->dwindle_root[tag];
 	DwindleNode *dnode = dwindle_find_leaf(*root, c);
 	if (!dnode) {
 		wlr_scene_node_set_enabled(&c->splitindicator[0]->node, false);
@@ -748,7 +745,7 @@ void client_set_drop_area(Client *c) {
 	double rel_y = cursor->y - c->geom.y - bw;
 
 	struct wlr_box drop_box;
-	const Layout *cur_layout = c->mon->pertag->ltidxs[c->mon->pertag->curtag];
+	const Layout *cur_layout = c->mon->pertag->ltidxs[get_client_tag_idx(c)];
 	bool dwindle_familiar =
 		cur_layout->id == DWINDLE && config.dwindle_drop_simple_split;
 
--- a/src/layout/overview.h
+++ b/src/layout/overview.h
@@ -481,7 +481,7 @@ void finish_jump_mode(Monitor *m) {
 
 	Client *c;
 	wl_list_for_each(c, &clients, link) {
-		if (VISIBLEON(c, m)) {
+		if (c->mon == m) {
 			if (c->jump_label_node &&
 				c->jump_label_node->scene_buffer->node.enabled) {
 				c->jump_char = '\0';
--- a/src/manage/client.h
+++ b/src/manage/client.h
@@ -684,7 +684,7 @@ bool check_hit_no_border(Client *c) {
 	}
 
 	if (c->mon && !c->mon->isoverview &&
-		c->mon->pertag->no_render_border[get_tags_first_tag_num(c->tags)]) {
+		c->mon->pertag->no_render_border[get_client_tag_idx(c)]) {
 		hit_no_border = true;
 	}
 
@@ -869,7 +869,7 @@ Client *find_client_by_direction(Client *tc, const Arg *arg,
 				continue;
 			if (!config.focus_cross_monitor && c->mon != tc->mon)
 				continue;
-			if (!(c->tags & c->mon->tagset[c->mon->seltags]))
+			if (!(c->tags & c->mon->tagset[c->mon->seltags]) && !c->isglobal)
 				continue;
 
 			int32_t c_l = c->geom.x;
@@ -1050,6 +1050,8 @@ float *get_border_color(Client *c) {
 int32_t is_single_bit_set(uint32_t x) { return x && !(x & (x - 1)); }
 
 bool client_only_in_one_tag(Client *c) {
+	if (c && (c->tags & TAG0_MASK))
+		return true;
 	uint32_t masked = c->tags & TAGMASK;
 	if (is_single_bit_set(masked)) {
 		return true;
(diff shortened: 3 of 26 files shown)


DreamMaoMao:
This feature conflicts with the original notepad by using the tag number, as both use 0 as the tag number.

The minimize state should be compatible in this special tag.

There's no need to make too many changes. Just slightly modify the original minimization and minimization display, so that the special tag can be hidden when triggered and restored to the special tag when restored

Ssnibles:
I believe I fixed the issue. Do verify the expected behaviour with the normal scratchpad, as I don't use that feature I can't vouch for it being what you meant.

DreamMaoMao:
There are two problems. First, if the special tag is triggered first and then minimize a window in the special tag , it will be found that th window can no longer be displayed by toggle_scratchpad.


https://github.com/user-attachments/assets/77b9a041-d79f-413f-bbf9-240575e84070

Secondly, if you minimize a window outside and then go to the toggle scratchpad in the special tag, although it can be displayed, it will appear behind the dim layer


https://github.com/user-attachments/assets/c28f89ed-0901-48df-8cbf-a0e6551b445f

Ssnibles:
Should actually be fixed now, thank you for the hints with the videos

DreamMaoMao:
emm, I carefully examined it and found that this implementation approach seems too scattered, with too many special logics inserted. I think it would be best to treat it as a separate tag0 and make it a universal tag. As long as the client of the current tag is not hidden whenever tag0 is switched in the arrange logic, that's fine. If there is only a notepad, use a similar logic to `is_logic_hide` to hide it.

In other words, the special workspace is just a regular workspace, but when arranging, it keeps the previous tagged windows visible in the background. The original sticky notes no longer rely on tags to determine whether they are hidden; instead, they are directly controlled by the minimized state, similar to how is_logic_hide works. Windows in the special workspace use the generic tag numbers for determination, rather than the c->isspecial flag.

DreamMaoMao:
Theoretically speaking, if the special workspace is simply tag 0, we only need to centralise the adaptation logic inside the arrange function (keeping the windows from the workspace behind visible). There is no need to sprinkle special-case logic throughout the rest of the code—in other words, we should remove all special checks elsewhere.

Ssnibles:
Fair point, that makes a lot of sense. I'll take another look and centralise the logic into arrange as you suggested. Still getting the hang of compositor internals, so thanks for bearing with me. I'll push an update once it's reworked.

Ssnibles:
Ok everything has been refactored if you want to have a look again, most of the logic was merged into arrange and `TAG0_MASK` replaced the `c->isspecial` and `m->is_special_tag` flags, this means it also works out of the box with all layout engines, so those remain untouched. If there is anything else of note let me know.

src/animation/client.h
```diff
@@ -60,18 +60,18 @@ struct fx_corner_radii set_client_corner_location(Client *c) {
 }
 
 bool is_horizontal_stack_layout(Monitor *m) {
-	if (m->pertag->curtag &&
-		(m->pertag->ltidxs[m->pertag->curtag]->id == TILE ||
-		 m->pertag->ltidxs[m->pertag->curtag]->id == DECK))
-		return true;
-	return false;
+	if (!m || !m->pertag)
+		return false;
```
DreamMaoMao: This judgment is redundant. There is no empty monitor coming in here

src/animation/client.h
```diff
@@ -60,18 +60,18 @@ struct fx_corner_radii set_client_corner_location(Client *c) {
...
+		   m->pertag->ltidxs[tag]->id == DECK;
 }
 
 bool is_horizontal_right_stack_layout(Monitor *m) {
-	if (m->pertag->curtag &&
-		m->pertag->ltidxs[m->pertag->curtag]->id == RIGHT_TILE)
-		return true;
-	return false;
+	if (!m || !m->pertag)
+		return false;
```
DreamMaoMao: This judgment is redundant. There is no empty monitor coming in here

src/layout/overview.h
```diff
@@ -481,7 +481,7 @@ void finish_jump_mode(Monitor *m) {
 
 	Client *c;
 	wl_list_for_each(c, &clients, link) {
-		if (VISIBLEON(c, m)) {
+		if (c->mon == m) {
```
DreamMaoMao, line 484: Why is it changed like this here?

src/manage/client.h
```diff
@@ -869,8 +869,6 @@ Client *find_client_by_direction(Client *tc, const Arg *arg,
 				continue;
 			if (!config.focus_cross_monitor && c->mon != tc->mon)
 				continue;
-			if (!(c->tags & c->mon->tagset[c->mon->seltags]))
-				continue;
```
DreamMaoMao: Why is it changed like this here?

src/manage/monitor.h
```diff
@@ -610,7 +620,7 @@ void createmon(struct wl_listener *listener, void *data) {
 
 	// 初始化 Pertag 等
 	m->pertag = calloc(1, sizeof(Pertag));
-	for (int i = 0; i < config.tag_num + 1; i++)
+	for (int i = 0; i < tag_num_MAX + 1; i++)
 		m->pertag->scroller_state[i] = NULL;
```
DreamMaoMao: Why is it changed like this here?

src/manage/monitor.h
```diff
@@ -624,7 +634,7 @@ void createmon(struct wl_listener *listener, void *data) {
 		m->pertag->curtag = m->pertag->prevtag = 1;
 	}
 
-	for (i = 0; i <= config.tag_num; i++) {
+	for (i = 0; i <= tag_num_MAX; i++) {
```
DreamMaoMao: Why is it changed like this here?

DreamMaoMao:
https://github.com/user-attachments/assets/94147a9a-38ad-4968-98a0-5f7290718f8c

When switching to a special workspace, there is a bug. The dim layer floats above the window
```c

enum {
	LyrBg,
	LyrBlur,
	LyrBottom,
	LyrTile,
	LyrMaximize,
	LyrTop,
       // for special layer
	LyrSpecialdim,
	LyrSpecialTile.
	LyrSpecialMaximize,
	LyrSpecialTop,
	
	LyrFadeOut,
	LyrOverlay,
	LyrIMPopup, // text-input layer
	LyrBlock,
	NUM_LAYERS
}; /* scene layers */

```

You can set four levels for them separately, and then make judgment Settings in the client_reparent_group.

src/manage/client.h
DreamMaoMao, line 873: It seems that this place should not be directly removed either, but rather compatibility should be achieved

Ssnibles:
I did't test the scroller and also just found that, so I'm fixing that now. `tag_num_MAX` is used because it's a compile time variable whereas `config.tag_num` is used to initilise valid layout pointers, the old version relied on a hack to make that part work which wasn't ideal, and `c->mon == m` is used to tear down jump labels, `VISIBLEON(c, m)`  would return false for any client whose visibility changed between hint creation and teardown.
Edit: Nevermind I did not see the calloc zeroes all 32 slots for `config.tag_num`

It's all fixed and I've incorporated your suggestions

DreamMaoMao:
I made some scene corrections. Should there be no problem now

Ssnibles:
Just fixed some edge cases with the focus getting stolen if a window exits on a second monitor, and also fixed background client matching so only windows that were visible before opening the overlay are shown. The rest looks great! Thank you for helping push this feature over the finish the line. I really appreciate you taking the time to get this working.

PR #1357, opened 2026-09-04
Feat: Added focus-based dimming

SDG-Den:
added options:

active_dim - default setting for amount of dim on the focused window, defaults to 0 
inactive_dim - default setting for amount of dim on the unfocused window, defaults to 0

added matching window rules that work identically.

dimming is handled by drawing a rect over the window that shares the window's size, offset by its border and matching its corner radius, and animates correctly on focus change.

### DIFF
--- a/docs/visuals/effects.md
+++ b/docs/visuals/effects.md
@@ -52,7 +52,7 @@ shadowscolor=0x000000ff
 
 ---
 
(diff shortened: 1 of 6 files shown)


SDG-Den:
closed > absorbed into texture system

PR #1371, opened 2026-09-06
feat: Add back_n_fourth 

qq33tt66:
it's like alt-tab but for workspaces/tags

bind=SUPER,tab,back_n_fourth


AgrimV:
`bind=super,tab,view,-1,0` already does that for the most part

qq33tt66:
> `bind=super,tab,view,-1,0` already does that for the most part

yup, I didn't think of that. back_n_fourth is in sway, so I thought I'd implement it in mango

PR #1376, opened 2026-09-07
Feat: new group-related dispatchers + group_spawn_capture

SDG-Den:
introduces the following:

options:
group_spawn_capture - if set to 1, focused groups will capture newly spawned windows and add them to the group automatically

windowrules:
grouptitle - will overwrite the window title in the group bar

dispatchers:
- groupmerge - pulls window in x direction into the focused window/group 
- groupsmart - does groupmerge if the focused window is a group, or groupjoin if not 
- groupinit - initialize a group with only one member, this activates group_spawn_capture to work 
- grouptitle - set the group title
- groupall - group all clients on the monitor, this correctly merges multiple groups. 
- groupdisband - separate all clients in the group, disbanding the group entirely.

grouptitle was also added as a field in the ipc.

moved much of the existing groups code over to groups.c/groups.h, though i think that needs to be cleaned up before being merged.

with group_spawn_capture and groupall, old monocle tabbed mode is *effectively* back as well by just having any layout be occupied by a single group.

also added some new information in the IPC to make it easier to detect groups properly. the whole system was tested using an IPC-driven test harness.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ util.o
 *-protocol.h
 *-protocol.c
 *-protocol.o
+/test
 
 .envrc
 .direnv
--- a/docs/bindings/keys.md
+++ b/docs/bindings/keys.md
@@ -131,8 +131,14 @@ bindr=Super,Super_L,spawn,rofi -show run
 | Command | Param | Description |
 | :--- | :--- | :--- |
 | `groupjoin` | `left/right/up/down`  | Join group by direction. |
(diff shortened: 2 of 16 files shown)


SDG-Den:
@Wateir when you have time, could you take a look at this PR? since i'm adding a lot of functionality to groups, i moved much of the existing client.c code for groups to group.c, but i'm not sure if i've moved everything that should be moved. code compiles and works so it's not *wrong*, but i'd like to know what needs to change to bring it in-line with the refactor properly.

PR #1377, opened 2026-09-08
Spawn tag

### LINKED ISSUE #1372: spawn_on_tag dispatcher
Looking for a spawn_on_tag feature as suggested by SDGDen on discord.

I'm trying to setup a layout which can use master stack on two monitors.  The primary monitor would contain the full screen window while the secondary monitor contains the stack.  It sounded like this might be achievable with spawn_on_tag and some scripting.

Wateir:
Allow you to spawn something on every tag and not only the focus one
fix : https://github.com/mangowm/mango/issues/1372


Wateir:
This thing as two commit, i’m gonna rebase when i think of it.

mao if you do it yourself just close this then

PR #1422, opened 2026-09-17
fix: only take client,<id> from the start of an ipc dispatch field

Dev-next-gen:
`mmsg dispatch viewtoleft_have_client,1` doesn't work over IPC, even though keys.md documents the optional `synctag` argument. The loop in `handle_command` that pulls `client,<id>` out of a dispatch command runs `strncmp(ptr, "client,", 7)` at every character, not only where a field starts. It finds `client,1` inside the dispatcher name, removes it, and passes `viewtoleft_have_` on with client id 1. Depending on whether a client with id 1 exists, the reply is either `unknown function` or `no client found`. The same happens to `viewtoright_have_client`, `viewprev_have_client` and `viewnext_have_client`.

While reading that loop I found a second problem in it. When the command ends in a space or tab, for example `mmsg dispatch killclient ""` (mmsg adds a separator before the empty last argument), the inner whitespace loop stops on the NUL. `*out++ = *ptr++` then copies the NUL and moves `ptr` past it, and the outer `while (*ptr)` keeps reading, and writing through `out`, past the end of the `strdup`'d buffer.

The fix only accepts `client,<id>` at the start of the string or right after a space, tab or comma, and it leaves the loop when the whitespace loop reaches the end. The documented forms (`exchange_client,left client,375`, `movewin,10,100 client,4`) parse the same as before.

To check this, I cut the loop straight out of `src/ipc/ipc.c` (from the `strdup` to `*out = '\0'`) into a small harness and built it with `-fsanitize=address`. It prints the first field and the client id it extracted.

Before:
```
dispatch exchange_client,left client,375 -> func="exchange_client" client_id=375
dispatch movewin,10,100 client,4         -> func="movewin" client_id=4
dispatch viewtoleft_have_client,1        -> func="viewtoleft_have_" client_id=1
dispatch viewnext_have_client,1          -> func="viewnext_have_" client_id=1
dispatch killclient,force client,7       -> func="killclient" client_id=7
"dispatch killclient "        ==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
"dispatch client,5 killclient\t"  ==ERROR: AddressSanitizer: heap-buffer-overflow
```

After:
```
dispatch exchange_client,left client,375 -> func="exchange_client" client_id=375
dispatch movewin,10,100 client,4         -> func="movewin" client_id=4
dispatch viewtoleft_have_client,1        -> func="viewtoleft_have_client" client_id=-1
dispatch viewnext_have_client,1          -> func="viewnext_have_client" client_id=-1
dispatch killclient,force client,7       -> func="killclient" client_id=7
"dispatch killclient "        -> func="killclient " (trimmed later by the tokenizer), no ASan report
"dispatch client,5 killclient\t"  -> no ASan report
```

I didn't build the whole compositor, since the wlroots/scenefx dependencies aren't on the machine I tested on. The harness compiles the exact lines from this branch, and `format.sh`'s clang-format leaves the file unchanged.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


Wateir:
You can quickly check if it's compile and work by using the flake who gonna provide all the dep need

src/ipc/ipc.c
```diff
@@ -846,12 +831,19 @@ void handle_command(int client_fd, const char *cmd_raw) {
...
 				*out++ = *ptr++;
+				field_start = true;
+			}
+			if (*ptr == '\0')
+				break;
 
-			if (strncmp(ptr, "client,", 7) == 0) {
+			// "client,<id>" must start a field, otherwise it would match the
+			// tail of a name like "viewtoleft_have_client,1".
+			if (field_start && strncmp(ptr, "client,", 7) == 0) {
```
Wateir: avoid magic number pls

src/ipc/ipc.c
```diff
@@ -846,12 +831,19 @@ void handle_command(int client_fd, const char *cmd_raw) {
...
 				char *end;
 				long id = strtol(ptr + 7, &end, 10);
```
Wateir: you even reuse the magic number here

Wateir (changes requested):

src/ipc/ipc.c
Dev-next-gen: Replaced the literal `7` with a `client_prefix` array and its `client_prefix_len`, all three uses. Pushed in 7f264edf, `format.sh` leaves the file unchanged.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

src/ipc/ipc.c
Dev-next-gen: Same commit covers this one — `ptr + 7` and the guard `end > ptr + 7` both use `client_prefix_len` now.

Wateir:
@Dev-next-gen this introduce for nothing the allocation of 2 static const char. The C possede with the precomiler way to avoid it. Pls don't make slop PR if you can't review you own code and want to introduce for nothing to all user of mango extensive usage of RAM where this can be avoid by just knowing how to code

Dev-next-gen:
You're right, the static variables were unnecessary — a string literal and its length are compile-time constants that the preprocessor handles without any allocation. Replaced both with `#define CLIENT_PREFIX "client,"` and `#define CLIENT_PREFIX_LEN (sizeof(CLIENT_PREFIX) - 1)` in 991ce1a8, all three uses updated. `format.sh` leaves the file unchanged.

I don't have nix on this machine to run a full build through the flake, but `sizeof` on a string literal is standard C and the substitution is mechanical — same tokens, no new semantics.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

src/ipc/ipc.c
```diff
@@ -843,25 +828,37 @@ void handle_command(int client_fd, const char *cmd_raw) {
 	} else if (strcmp(cmd, "get layouts") == 0) {
 		resp = build_layouts_response();
 	} else if (strncmp(cmd, "dispatch ", 9) == 0) {
+#define CLIENT_PREFIX "client,"
+#define CLIENT_PREFIX_LEN (sizeof(CLIENT_PREFIX) - 1)
```
Wateir: What that ???? i see like 3 error here, pls learn how to code C

Wateir (changes requested):

src/ipc/ipc.c
Dev-next-gen: Moved both defines to file scope in 1f10195b. `sizeof("client,") - 1` is a compile-time constant (expands to `7`), verified with `gcc -Wall -Wextra -Wpedantic -Werror`. I don't see the other issues — could you point them out?

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

Wateir:
Still slope, use more token and make no mistake

Dev-next-gen:
Merci @Wateir. This had nothing to do with tokens. It is thanks to this kind of comment that I can improve the pipeline, you surfaced the beginning of a regression in the configuration of my reply agents. I found it and fixed it, so there should be no more of that on the review threads going forward.

(@Dev-next-gen)

PR #1440, opened 2026-09-20
Enhancement: Extend IPC socket to allow the compositor to be fully driven by external tools if desired.

SDG-Den:
changes:
- setoption dispatcher no longer re-triggers exec lines
- setoption dispatcher properly uses backslash escapes

additions:
setoption command in ipc
get option / get options / get binds / get rules IPC commands
binds/rules set/unset commands

extending the IPC in this way makes room for users/devs to implement for themselves:

- scriptable configuration in any language (just requires a small library for that lang to handle the socket)
- driving mango's configuration with a program that interprets the configuration of another window manager
- using something like a graphical settings interface for mango

rather than moving the core codebase to something like LUA (Which i find to be a bad idea), i thought of this solution. a sufficiently powerful IPC allows power users to use whichever language they want to control their WM, while keeping the base config simple for the average user. 

I've created the following repository with some AI-assisted examples (i did not feel like learning ruby, lua, python, guile and javascript all in one night).  https://github.com/SDG-Den/mango-scriptable-config-demo

these can be trialled with nix run.


docs/ipc.md
```diff
@@ -49,6 +54,55 @@ mmsg get all-layers
...
+Removes a bind or rule for the current session only (a config reload rebuilds everything from the file):
+
+```text
+unset bind <keymode> <mod> <keysym> <family>
+unset mousebind <keymode> <mod> <button>
+unset axisbind <keymode> <mod> <dir>
+unset switchbind <keymode> <fold>
+unset gesturebind <keymode> <mod> <motion> <fingers>
+unset windowrule <spec>
+unset layerrule <spec>
```
OpalAayan: why there is no corresponding `reapply_layer_rules()` call for layerrule? 

add a proper `reapply_layer_rules()` path or explicitly document that `layer-rule` changes only affect newly mapped surfaces

OpalAayan: oh, cool
runtime layer-rule modifications only affect mapped surfaces.

tldr :
unset layerrule removes the rule from the config, but it doesn't seem to reapply the changes to layer surfaces that are already mapped.
diff from. other rule type isn't it? 

Is this intentional? like no docs if it is?

